PR #246, opened 2014-12-08
Add space user role assignation for current user.

o-orand:
Hello Scott
As space creation does not handle user role, it has to be done after. So once a space is created, current logged user can be assign to it. Assigned user roles are identical to CF CLI ones, ie Manager, Developer
and Auditor.

Only assign and list are implemented.

Related to PR #238


scottfrederick:
@o-orand Sorry for the delay in reviewing. Looks good, merged. Thanks!

scottfrederick:
@o-orand I renamed some methods from this PR for consistency with the rest of the API (most methods that retrieve data start with `get` whether they return a single entity or a collection of entities).

Changes are in https://github.com/cloudfoundry/cf-java-client/commit/a435a447495638c1933c7a784bf07d207a0a8451

o-orand:
@scottfrederick. Thanks for feedback and the merge.

Issue #214, opened 2014-09-19
CF Client throws NPE when using OAuthToken without refresh and no expiration


kdvolder:
Version: 1.0.3

If you use a OAuthToken object like this:

         new DefaultOAuth2AccessToken("...")

You get a token that has no expiration and no refresh token. The way I read the docs this should be ok, you just don't get to refresh that token.

When I put this token in the CF client the client tries to refresh the token when I try connect logregator listener. It sees time until expiration as '0' and thinks its time to refresh. But since there's no refresh token it throws an NPE.

java.lang.NullPointerException
	at org.springframework.security.oauth2.client.token.grant.password.ResourceOwnerPasswordAccessTokenProvider.refreshAccessToken(ResourceOwnerPasswordAccessTokenProvider.java:39)
	at org.cloudfoundry.client.lib.oauth2.OauthClient.refreshToken(OauthClient.java:124)
	at org.cloudfoundry.client.lib.oauth2.OauthClient.getToken(OauthClient.java:88)
	at org.cloudfoundry.client.lib.rest.CloudControllerClientImpl.getUserInfo(CloudControllerClientImpl.java:1722)
	at org.cloudfoundry.client.lib.rest.CloudControllerClientImpl.getInfo(CloudControllerClientImpl.java:534)
	at org.cloudfoundry.client.lib.rest.CloudControllerClientImpl.streamLoggregatorLogs(CloudControllerClientImpl.java:1631)
	at org.cloudfoundry.client.lib.rest.CloudControllerClientImpl.streamLogs(CloudControllerClientImpl.java:233)
	at org.cloudfoundry.client.lib.CloudFoundryClient.streamLogs(CloudFoundryClient.java:336)
	at org.eclipse.flux.cloudfoundry.deployment.service.CloudFoundryClientDelegate$DeployedApplicationLogListener.<init>(CloudFoundryClientDelegate.java:348)
	at org.eclipse.flux.cloudfoundry.deployment.service.CloudFoundryClientDelegate.addLogListener(CloudFoundryClientDelegate.java:186)
	at org.eclipse.flux.cloudfoundry.deployment.service.CloudFoundryClientDelegate$4.doRun(CloudFoundryClientDelegate.java:157)
	at org.eclipse.flux.cloudfoundry.deployment.service.CloudFoundryClientDelegate$4.doRun(CloudFoundryClientDelegate.java:1)
[...]

PR #259, opened 2015-02-12
Allow users to specify weather a token is refreshable or not

asaikali:
OAuth2 tokens obtain via implicit OAuth 2.0 authentication can not be refreshed,
and since there is no username / password passed in to the cloud credentials we
just need to use the provided token without trying to refresh.

Resolves issue https://github.com/cloudfoundry/cf-java-client/issues/214


scottfrederick:
@asaikali This looks really good Adib. My only request is to re-format the changed code to use tabs for indentation instead of spaces to match the formatting of the rest of the code. If you can make that change in another commit to the same fork, github will update the PR and I will merge it.

asaikali:
I have fixed the indent issues, please let me know if there is anything else need me to change.

asaikali:
I am currently working for Pivotal do I still need to sign the CLA?

Issue #263, opened 2015-02-26
No way to get env variables


andreaturli:
I'd like to have a method to get the env variables as from the CF web console.
`app.getEnv` and `app.getEnvAsMap` don't seem to work as expected.

What am I missing?

scottfrederick:
The current `getEnv()` and `getEnvAsMap()` methods only return user-provided environment variables (those set with `cf set-env` or the CC API equivalent). We have an open issue to expand those to support system env vars: #204.

andreaturli:
thanks @scottfrederick, do you think it is hard to solve #204?

andreaturli:
Alternatively @scottfrederick is there a way to discover the VCAP_SERVICES properties exposed by a service?

scottfrederick:
The fix for this isn't hard, it just needs to be prioritized. So far it has just been a "nice to have" feature. 

Currently, the credentials for a service instance are only provided when the service instance is bound to an application, so querying the application's environment is the only way to get that information.

andreaturli:
@scottfrederick, please consider my PR #264 as a potential fix.

scottfrederick:
Fixed via #264

PR #264, opened 2015-02-26
Fix issue #263

andreaturli:
add support for application environment variables

I've completed your CLA Agreement for individuals and just sent at contributors@cloudfoundry.org

### DIFF
--- a/cloudfoundry-client-lib/src/main/java/org/cloudfoundry/client/lib/CloudFoundryClient.java
+++ b/cloudfoundry-client-lib/src/main/java/org/cloudfoundry/client/lib/CloudFoundryClient.java
@@ -223,12 +223,16 @@ public class CloudFoundryClient implements CloudFoundryOperations {
 		return cc.getApplication(appName);
 	}
 
-   public Map<String, Object> getApplicationEnvironment(UUID appGuid) {
-      return cc.getApplicationEnvironment(appGuid);
-   }
-
 	public CloudApplication getApplication(UUID appGuid) {
 		return cc.getApplication(appGuid);
+	}
+
+	public Map<String, Object> getApplicationEnvironment(UUID appGuid) {
+		return cc.getApplicationEnvironment(appGuid);
+	}
+
+   	public Map<String, Object> getApplicationEnvironment(String appName) {
+    	return cc.getApplicationEnvironment(appName);
 	}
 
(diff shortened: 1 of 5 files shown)


jbayer:
@scottfrederick we've received the CLA from @andreaturli

thanks for your contribution!

scottfrederick:
Thanks @andreaturli, this looks good! 

One request on formatting - please use tabs for indentation on new code to remain consistent with existing code, and avoid re-formatting existing code to replace tabs with spaces. 

Also, most methods that act on applications take the app name as the parameter. To be consistent, we should have a `getAppEnvironment` method that takes the app name, in addition to the one that takes a GUID. 

If you can make these changes and push them to the same fork and branch, the PR will be updated and I will merge it. 

Thanks again for the contribution.

andreaturli:
thanks @scottfrederick I've addressed your comments and added `getApplicationEnvironment(String appName)` with its test. I've also fixed the formatting, I think.

Thanks for your review

If that will be merged, is there a maven repository where I can get the SNAPSHOT?

scottfrederick:
Well, I'm afraid the formatting is worse now :-) Look at this view:
https://github.com/cloudfoundry/cf-java-client/pull/264/files. Now there
are a bunch of unrelated whitespace changes and re-organizing of import
statements that weren't there before, which makes it harder to see what the
real changes were. And it appears the added code still isn't indented the
same as the surrounding code:
https://github.com/cloudfoundry/cf-java-client/pull/264/files#diff-604d152f5bfadf46ec23ac5ce642ed3eR210
.

Can you take one more try at the formatting, trying to fix only the code
that was added or changed?

Unfortunately we don't publish snapshots anywhere right now. I'd like to
publish snapshots and milestones to repo.spring.io but that's another to-do
item.

andreaturli:
Ops sorry, I think now it is better even if my IDE decided to remove some trailing spaces inside the comments. Is it acceptable now?

scottfrederick:
Yeah, that's better. The extra formatting still creates a lot of noise in
the diff that makes it hard to see what the real changes are. It's best not
to let your IDE do re-formatting of existing code when working with code
like this that has lots of history to it. I'll merge this as-is.

andreaturli:
Thanks @scottfrederick, sorry again for the formatting issue.

Is there a roadmap for the new release of `cf-java-client/cloudfoundry-client-lib`?

Issue #271, opened 2015-03-24
getSpaceDevelopers(String spaceName) returns NPE in version 1.1.1


AndreasHeix:
Hi,
not sure if this is related to https://github.com/cloudfoundry/cf-java-client/issues/269.

Here's my code:
		CloudCredentials credentials = new CloudCredentials(user, password);
		client = new CloudFoundryClient(credentials, getTargetURL(target), true);
		client.login();
		//try to force more metadata, see https://github.com/cloudfoundry/cf-java-client/issues/269
		CloudSpace space = client.getSpace("foo");
		space.getOrganization();
		//now back to the call I really want to do
		List<UUID> developers = client.getSpaceDevelopers("foo");

However, I am getting an NPE in CloudControllerClientImpl:

	private List<UUID> getSpaceUserGuids(String spaceName, String urlPath) {
		UUID orgGuid = sessionSpace.getOrganization().getMeta().getGuid();

because the sessionSpace is "null".

It looks a little bit difficult to me anyhow, as there can be several spaces with the same name in a CF landscape albeit in different orgs. So: Wouldn't I have to specify the org as well when I want to retrieve details about a space?

Thanks and regards
Andreas

scottfrederick:
@AndreasHeix That's a bug in `getSpaceUserGuids()`. Other `CloudFoundryClient` constructors allow you to pass a default org and space that operations can act on, which is what the `sessionSpace` field gets set from. `getSpaceUserGuids()` shouldn't assume that this field is always set.

I think you are right about specifying the org on this call. We can leave the current forms of the `getSpaceXxxx()` methods that just take a space name, and have them throw a more useful exception if the default org wasn't specified at construction. Then we can add another form of those methods that takes an org parameter to meet your use case. The same should be done with the `associateXxxxWithSpace()` methods.

scottfrederick:
Fixed via https://github.com/cloudfoundry/cf-java-client/commit/cca3986ba056bf1f79f4b73c594a0f02db2761f6 and https://github.com/cloudfoundry/cf-java-client/commit/26c32370b577d53b0123e77efd2ed09ee06bcd8c.

PR #280, opened 2015-05-11
Associate any organization user to a space.

### LINKED ISSUE #278: As CLI does, add any user to a space.
Client-lib should allow user with enough privilege to assign another user to a space.

For me it's not clear how can I get user guid from a username with client lib. 

The CLI is using UAA to retreive user guid
```
GET /Users?attributes=id,userName&filter=userName+Eq+%22aValidUsername%22 HTTP/1.1
Host: uaa.xxx.com
```
Once you've got the user guid , it's almost done.

CC API only provides a way to get all user using GET /v2/users for an organization (it's look like  the most appropriate filter). And search for the username in the returned list. Do you have other ideas ?

Regards,
 Olivier

o-orand:
Add userdGuid to  associatexxxxWithSpace method, if null, use current user.
Fix associatexxxxWithSpace orgName param used to search for space: if null use current org, else use orgName.
Add getOrganizationUsers to be able to retrieve all users of an organization indexed by username.
Change associatexxxxWithSpace method signatures introduced with #271 - getSpaceDevelopers(String spaceName) returns NPE in version 1.1.1

This closes #278 - As CLI does, add any user to a space.

### DIFF
--- a/cloudfoundry-client-lib/src/main/java/org/cloudfoundry/client/lib/CloudFoundryClient.java
+++ b/cloudfoundry-client-lib/src/main/java/org/cloudfoundry/client/lib/CloudFoundryClient.java
@@ -45,6 +45,7 @@ import org.cloudfoundry.client.lib.domain.CloudStack;
 import org.cloudfoundry.client.lib.domain.CrashesInfo;
 import org.cloudfoundry.client.lib.domain.InstancesInfo;
 import org.cloudfoundry.client.lib.domain.Staging;
+import org.cloudfoundry.client.lib.domain.CloudUser;
 import org.cloudfoundry.client.lib.rest.CloudControllerClient;
 import org.cloudfoundry.client.lib.rest.CloudControllerClientFactory;
 import org.springframework.security.oauth2.common.OAuth2AccessToken;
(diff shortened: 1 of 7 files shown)


cloudfoundry-client-lib/src/main/java/org/cloudfoundry/client/lib/CloudFoundryClient.java
```diff
@@ -25,26 +25,8 @@
 import java.util.UUID;
 
 import org.cloudfoundry.client.lib.archive.ApplicationArchive;
-import org.cloudfoundry.client.lib.domain.ApplicationLog;
-import org.cloudfoundry.client.lib.domain.ApplicationStats;
-import org.cloudfoundry.client.lib.domain.CloudApplication;
+import org.cloudfoundry.client.lib.domain.*;
```
scottfrederick: Please don't re-organize imports to use wildcards.

cloudfoundry-client-lib/src/main/java/org/cloudfoundry/client/lib/CloudFoundryOperations.java
```diff
@@ -16,6 +16,11 @@
 
 package org.cloudfoundry.client.lib;
 
+import org.cloudfoundry.client.lib.archive.ApplicationArchive;
+import org.cloudfoundry.client.lib.domain.*;
```
scottfrederick: Please don't re-organize imports to use wildcards.

cloudfoundry-client-lib/src/main/java/org/cloudfoundry/client/lib/CloudFoundryOperations.java
```diff
@@ -159,29 +141,32 @@
...
-	void associateDeveloperWithSpace(String orgName, String spaceName);
-
-	/**
-	 * Associate current user to the space managers role
-	 *
-	 * @param orgName name of the organization containing the space
-	 * @param spaceName name of the space
-	 */
-	void associateManagerWithSpace(String orgName, String spaceName);
+    /**
```
scottfrederick: These lines appear to be indented with spaces instead of tabs.

scottfrederick:
Thanks for the PR @o-orand. 

Adding the third `String userGuid` parameter to the `associateXxxxWithSpace` methods makes this a breaking API change, which we try to avoid with point releases. For backward compatibility, it would be better to leave the existing methods intact and add additional methods that have the third parameter.

o-orand:
@scottfrederick, thanks for feedback. I've updated the PR. I fixed formatting issues (IDE default...) and I restored ```associateXxxxWithSpace``` 2 params version to restore backward compatibility.

PR #286, opened 2015-06-09
Added methods to switch the current space. Furthermore overloaded signatures with CloudSpace.

devproof:
Added methods to switch the current space ("session space") to avoid a new instanciation and relogin of the client. Furthermore overloaded signatures with CloudSpace parameter in it to query concret results which can be outside of the session space.

### DIFF
--- a/cloudfoundry-client-lib/src/main/java/org/cloudfoundry/client/lib/CloudFoundryOperations.java
+++ b/cloudfoundry-client-lib/src/main/java/org/cloudfoundry/client/lib/CloudFoundryOperations.java
@@ -198,6 +198,15 @@ public interface CloudFoundryOperations {
 	CloudSpace getSpace(String spaceName);
 
 	/**
+	 * Get space name with the specified name.
+	 *
+	 * @param guid guid of the space
+	 * @return the cloud space
+	 */
+	CloudSpace getSpace(UUID guid);
+
+
+	/**
 	 * Delete a space with the specified name
 	 *
 	 * @param spaceName name of the space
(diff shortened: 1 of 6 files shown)


cloudfoundry-client-lib/src/main/java/org/cloudfoundry/client/lib/CloudFoundryOperations.java
```diff
@@ -25,25 +25,7 @@
...
-import org.cloudfoundry.client.lib.domain.CloudService;
-import org.cloudfoundry.client.lib.domain.CloudServiceBroker;
-import org.cloudfoundry.client.lib.domain.CloudServiceInstance;
-import org.cloudfoundry.client.lib.domain.CloudServiceOffering;
-import org.cloudfoundry.client.lib.domain.CloudSpace;
-import org.cloudfoundry.client.lib.domain.CloudStack;
-import org.cloudfoundry.client.lib.domain.CrashesInfo;
-import org.cloudfoundry.client.lib.domain.InstancesInfo;
-import org.cloudfoundry.client.lib.domain.Staging;
+import org.cloudfoundry.client.lib.domain.*;
```
scottfrederick: Please don't collapse or re-arrange imports.

scottfrederick:
@devproof Do you have a CLA on file, as specified [above](https://github.com/cloudfoundry/cf-java-client/pull/286#issuecomment-110258637)?

This change adds a large number of methods to an already large and monolithic API. Also, the changes allow passing a `CloudSpace` to control the space the library is acting on, but not `CloudOrg`, which seems uneven.

What is the use case for passing a `CloudSpace` to all these new methods, instead of just setting the current org and space and relying on that state like the library does today.

devproof:
@scottfrederick do I need it? I'm working for Pivotal Services (EMC).

devproof:
Is the CLA required for me? I'm working for Pivotal Services (EMC).

nebhale:
@devproof Thanks for your contribution. We won't be adding any new APIs in the V1 line of development. This is already covered as part of the V2 work being today today and can be tracked [here](https://www.pivotaltracker.com/story/show/106155428). I'm going to close this out but if you'd like to see it re-opened, please comment on the issue and I'll reopen it.

PR #291, opened 2015-06-30
Added the detected_start_command attribute to Staging.

davidehringer:
Added a field to the Staging class to exposed the command detected by the buildpack during staging.

This attribute was already being returned in the payload and didn't require any additional calls to the Cloud Controller.


nebhale:
Thanks for your contribution. We won't be adding any new APIs in the V1 line of development. This is already covered as part of the V2 work being done today. I'm going to close this out but if you'd like to see it re-opened, please comment on the issue and I'll reopen it.

PR #303, opened 2015-08-20
Fix mapping of a job's error details

WGautier:
As seen [in the API](http://apidocs.cloudfoundry.org/215/jobs/retrieve_job_with_known_failure.html), when retrieving the status of a failed job, the error is returned under `entity>error_details>*`. However, the CF java client tries to fetch `error_details>entity>*` instead.

This later causes a NullPointerException [here](https://github.com/cloudfoundry/cf-java-client/blob/7a8ab8c0770bde2854f798cbe97e6a90f8273a71/cloudfoundry-client-lib/src/main/java/org/cloudfoundry/client/lib/rest/CloudControllerClientImpl.java#L2148) instead of the correct CloudOperationException, and hides the true error message coming from the target.


glyn:
I copied the change across manually to a [new commit](https://github.com/cloudfoundry/cf-java-client/commit/8582ffbc37b24c07cfc5d9ecfb19c3a0ff580687) on the 1.x branch. Thanks for the fix!

PR #310, opened 2015-09-04
Add binding for org memory_usage API endpoint

jghiloni:
Add binding for the `memory_usage` API endpoint (available since v193) to allow folks to see how much RAM your org is using


nebhale:
@jghiloni Thanks for your contribution. We won't be adding any new APIs in the V1 line of development. This is already covered as part of the V2 work being today today and can be tracked [here](https://www.pivotaltracker.com/story/show/101426262). I'm going to close this out but if you'd like to see it re-opened, please comment on the issue and I'll reopen it.

PR #314, opened 2015-09-17
Adding deleteApplicationInstance(appName, index) to terminate specific instances of an application

bryancuster:
The method takes the application name and the instance number (zero
indexed). It calls DELETE /v2/apps/{guid}/instances/{index}.


nebhale:
@bryancuster Thanks for your contribution. We won't be adding any new APIs in the V1 line of development. This is already covered as part of the V2 work being today today and can be tracked [here](https://www.pivotaltracker.com/story/show/100988118). I'm going to close this out but if you'd like to see it re-opened, please comment on the issue and I'll reopen it.

PR #315, opened 2015-09-21
@muz allow command line overriding of the file attribute.

onekilo79:
Allow the overriding in the command line for the deployment file.  For CI and versioning.


onekilo79:
Not allowing default behaviour

PR #318, opened 2015-10-06
[gradle-plugin] overriding of the file attribute from the command line

onekilo79:
Allow deploy and push actions with an override from the command line with the gradle plugin.

E.G.: -PcfFile=a-new-file.jar

This throws a runtime exception if the override is used and it cannot be found.  

I did not want it to be invoked then silently fail to deploy a file that does not exist.  Currently it will then try to see if there is a jar to deploy and then the war.


nebhale:
@onekilo79 Is there a reason that this can't be solved via a variable in the use configuration rather than in the plugin itself?

nebhale:
Since there's been no response on this issue in a couple of weeks, I'm going to close it.  If you'd like to see it re-opened, please comment on the issue and I'll reopen it.

PR #325, opened 2015-11-03
FIX: Skip SSL Validation

antechrestos:
- No negation: if skipSslValidation is set to true then we need to trust the target host
- Use of Optional.ofNullable otherwise a Nullpointer will be raised when skipSslValidation is not provided

### DIFF
--- a/cloudfoundry-client-spring/src/main/lombok/org/cloudfoundry/client/spring/SpringCloudFoundryClient.java
+++ b/cloudfoundry-client-spring/src/main/lombok/org/cloudfoundry/client/spring/SpringCloudFoundryClient.java
@@ -107,7 +107,7 @@ public final class SpringCloudFoundryClient implements CloudFoundryClient {
                              String username, String password, RestOperations bootstrapRestOperations,
                              SslCertificateTruster sslCertificateTruster) {
 
-        if (!Optional.of(skipSslValidation).orElse(false)) {
+        if (!Optional.ofNullable(skipSslValidation).orElse(false)) {
             try {
                 sslCertificateTruster.trust(host, 443, 5, SECONDS);
             } catch (GeneralSecurityException | IOException e) {


nebhale:
@antechrestos You're definitely right that there's a bad handling of `null`, so the `ofNullable()` is warranted.  However, I don't think you can skip the negation.  You can see our [test cases proving that](https://github.com/cloudfoundry/cf-java-client/blob/master/cloudfoundry-client-spring/src/test/java/org/cloudfoundry/client/spring/SpringCloudFoundryClientTest.java#L52-L85).  I suspect that if your run the test suite against your changes, it'd fail.

antechrestos:
@nebhale Hi,
Thank you for your quick reply.
On our server we need to skip ssl validation (the certificate is self signed):
- if I provide skipSslValidation with a true value (telling him I need to skip the ssl) it fails.
- If I provide false it works.
It does the contrary that is expected given the variable name, doesn't it?

nebhale:
Are you sure that you're using latest `master`?  Specifically, if you could write a test case in the [`SpringCloudFactoryClientTests`](https://github.com/cloudfoundry/cf-java-client/blob/master/cloudfoundry-client-spring/src/test/java/org/cloudfoundry/client/spring/SpringCloudFoundryClientTest.java) that demonstrates the problem and your fix, that would go a long way towards helping me understand the problem.

antechrestos:
@nebhale : My apologies!
I understood the way you made it by reading the test you pointed me to. In my head skipping verification meant "trusting" it and therefore calling sslCertificateTruster.trust.

In previous version of client; on our cloudfoundry instance, I had to set skip ssl verification. Now I have to set it to false and provide my own SslCertificateTruster to be clear.

I updated the file with the return of the negation. ;)

nebhale:
Actually, I wouldn't be calling a constructor at all.  Instead, I think you want to do:

```
SpringCloudFoundryClient.builder()
                .host("test-host")
                .username("test-username")
                .password("test-password")
                .skipSslValidation(true)
                .build();
```

Why don't you just close out this PR and open an issue for the `null` problem.  I'll just go ahead and do the work since it's so trivial.

antechrestos:
Ok I will open an issue.
I did use the builder (not constructor sorry)
```java
client = SpringCloudFoundryClient.builder()
                    .host(clientConfiguration.getTargetEndpointHost())
                    .clientId(clientConfiguration.getClientId())
                    .clientSecret(clientConfiguration.getClientSecret())
                    .skipSslValidation(true)
                    .username(clientConfiguration.getUsername())
                    .password(clientConfiguration.getPassword()).build();
```
And I receive:
```
org.springframework.web.client.ResourceAccessException: I/O error on GET request for "https://api.nd-cfapi.itn.ftgroup/info":sun.security.validator.ValidatorException: PKIX path building failed: sun.security.provider.certpath.SunCertPathBuilderException: unable to find valid certification path to requested target; nested exception is javax.net.ssl.SSLHandshakeException: sun.security.validator.ValidatorException: PKIX path building failed: sun.security.provider.certpath.SunCertPathBuilderException: unable to find valid certification path to requested target
...
Caused by: javax.net.ssl.SSLHandshakeException: sun.security.validator.ValidatorException: PKIX path building failed: sun.security.provider.certpath.SunCertPathBuilderException: unable to find valid certification path to requested target
```
That is why I told you that I misunderstood the behavior of the skipSslValidation... I thought it was for the underlying ssl verification...

If I change to 
```java
client = SpringCloudFoundryClient.builder()
                    .host(clientConfiguration.getTargetEndpointHost())
                    .clientId(clientConfiguration.getClientId())
                    .clientSecret(clientConfiguration.getClientSecret())
                    .skipSslValidation(false)
                    .username(clientConfiguration.getUsername())
                    .password(clientConfiguration.getPassword()).build();
```
It works like a charm since it invoke trust on the target endpoint for 5 seconds...

nebhale:
@antechrestos I've been totally wrong this entire time, and you've been totally right!  I had the behavior of the `Truster` completely reversed (which probably means that we need to add some user documentation to keep us honest).  The test is backwards which is what's causing the confusion.  I'm going to reopen this issue and merge your change to get credit.  I'll update it a bit so that it contains passing tests as well.

nebhale:
Resolved by e015bff0665a16bf3e9bcae47b88b256055cc2c0

antechrestos:
Thanks :)

PR #338, opened 2016-01-07
Add the create service binding for an application

antechrestos:
I implemented the [create service binding operation](https://www.pivotaltracker.com/projects/816799/stories/101451518) for application v2.
I put it in the applicationv2 operation interface since __remove service binding__ was already here. I was tempted to create an interface that is dedicated to service binding and put in it the create operation, but also move the delete in it but I don't know very well if that is what is expected.
As usual I am opened to suggestions.

Note: to compile I had to revert dependency of reactor back to 2.1 since I had some compilation failures ```SpringLoggregatorClient``` for instance (some class such as ```Publishers``` are not present anymore)


nebhale:
You've missed on the location for the API.  This is not part of the Application API (`/v2/service_bindings`) so it'll need to change both location and names.  Once you're made the changes, remember to force push your branch under the same name so that it reuses this PR.

antechrestos:
@nebhale : OK

nebhale:
@antechrestos We're only adding a single API per commit (discrete units of work), so you'll need to break this into to PRs for next time.  I've gone ahead and done it this time (and updated the commit messages to be accurate).

see: a91eddc478c2a03689a2f28d7ca6a74ce480631a
see: 6ddf7c49c6125e4673a62bb907daaa52bf82afd5

antechrestos:
@nebhale : Ok!
The title of [this](https://github.com/cloudfoundry/cf-java-client/commit/a91eddc478c2a03689a2f28d7ca6a74ce480631a) commit is not so clear :smile:

PR #346, opened 2016-01-12
List Service Bindings for a Service Instance

antechrestos:
This change adds the API for List Service Binding for a Service Instance (GET /v2/service_instances/:guid/service_bindings) according to [this story](https://www.pivotaltracker.com/projects/816799/stories/101451540)


nebhale:
@antechrestos Minor polishing required, but another great submission.  You even caught the doc error that `service_instance_guid` isn't a valid filter.

PR #353, opened 2016-01-15
Create Service Broker

antechrestos:
This change adds the API for Create Service Broker (```POST /v2/service_brokers```)
Described in [this story](https://www.pivotaltracker.com/projects/816799/stories/101451526)

PR #355, opened 2016-01-16
Renaming

antechrestos:
As mentioned in #353 resource files were not well named.
These modifications report the advise on work that has been done before.
As far as I understood:
- since in both service binding and instance there is only one ```POST``` without any id as path parameter, ti will be  named ```POST_<request/response>.json```
- only one ```PUT``` operation in service binding, that takes an id in its path. Hence ```PUT_{id}_<request/response>.json```


nebhale:
@antechrestos Your understanding is correct.  The rule of thumb is that they look like the request they map to, with `_request` or `_response` appended to them.  So `POST /v2/service_instances` goes to `/v2/service_instances/POST_request/response.json`, `PUT /v2/service_instances/:id` goes to `v2/service_instances/PUT_{id}_request/response.json`, and `PUT /v2/organizations/:id/auditors` goes to `/v2/organizations/PUT_{id}_auditors_request/response.json`.

PR #357, opened 2016-01-18
Bind a Service Instance To A Route

antechrestos:
This change adds the API for Bind a Service Instance To a Route (```PUT /v2/service_instances/:service_instance_guid/routes/:route_guid```)
We know that it is experimental, but @gberche-orange  mentionned me that we may need it in autosleep development.


gberche-orange:
If this isn't clear, this corresponds to the route services support in CC API, with the [binding of a "route service" instance to a route](http://apidocs.cloudfoundry.org/227/service_instances/binding_a_service_instance_to_a_route_%28experimental%29.html). 

More general route services pointer into http://docs.cloudfoundry.org/services/route-services.html and [design spec](https://docs.google.com/document/d/1SfwaQ1hnngfopXC_Q24cT6lbo0yFwvbAbPcCPEHeNPY/edit?usp=sharing)

PR #358, opened 2016-01-19
List Service Brokers

antechrestos:
This change adds the API for List Service Broker (```GET /v2/service_brokers```)
Described by [this story](#101451530)


nebhale:
Absolutely perfect!  Zero changes.

PR #360, opened 2016-01-20
List Service Instance for a Service Plan

antechrestos:
- Init Service Plan API
- adds the API for List Service Instance for a Service Plan (GET /v2/service_plans/:guid/service_instances)
According to [this story](https://www.pivotaltracker.com/projects/816799/stories/101451574)
[#101451574]

_According to the documentation: _Valid filters: name, space_guid, service_plan_guid, service_binding_guid, gateway_name, organization_guid, service_key_guid_

- I willingly removed ```service_plan_guid``` filter because it is in the request...Did I miss something?
- Is ```organization_guid``` a valid filter (not an attribute of the service instance)?

See below


antechrestos:
After some tests with my cf-python-client I got the result:
```InvalidStatusCode: 400 : {"code": 10005, "error_code": "CF-BadQueryParameter", "description": "The query parameter is invalid: organization_guid"}```

nebhale:
@antechrestos You've stumbled upon the fact that the API documentation is often incorrect.  We open issues for each and every one of these problems in the hopes that they'll one day be fixed.

PR #362, opened 2016-01-21
List Service Plans

antechrestos:
This change adds the API for List Service Plans (```GET /v2/service_plans```)
Described in [This story](#101451576)

### DIFF
--- a/cloudfoundry-client/src/main/lombok/org/cloudfoundry/client/v2/serviceplans/ListServicePlansRequest.java
+++ b/cloudfoundry-client/src/main/lombok/org/cloudfoundry/client/v2/serviceplans/ListServicePlansRequest.java
@@ -1,0 +1,95 @@
+/*
+ * Copyright 2013-2016 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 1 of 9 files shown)


cloudfoundry-client/src/main/lombok/org/cloudfoundry/client/v2/serviceplans/ServicePlanEntity.java
```diff
@@ -0,0 +1,133 @@
...
+     */
+    private final String uniqueId;
+
+    /**
+     * The visible flag
+     *
+     * @param public the visible flag
+     * @return the visible flag
+     */
+    private final boolean visible;
```
nebhale: @antechrestos Good naming choice.

cloudfoundry-client/src/main/lombok/org/cloudfoundry/client/v2/serviceplans/ListServicePlansRequest.java
```diff
@@ -0,0 +1,95 @@
...
+@ToString(callSuper = true)
+public final class ListServicePlansRequest extends PaginatedRequest implements Validatable {
+
+    /**
+     * The active flag
+     *
+     * @param active the active flag
+     * @return the active flag
+     */
+    @Getter(onMethod = @__(@QueryParameter("active")))
```
nebhale: @antechrestos Until documented (or a bug is opened against the documentation) we assume that all of the filters are `@InFilterParameter`.  If it's not, it'll get caught when we integration test the implementations.

cloudfoundry-client/src/main/lombok/org/cloudfoundry/client/v2/serviceplans/ServicePlanEntity.java
```diff
@@ -0,0 +1,133 @@
...
+@Data
+public final class ServicePlanEntity {
+
+    /**
+     * The visible flag
+     *
+     * @param public the visible flag
+     * @return the visible flag
+     */
+    private final boolean active;
```
nebhale: @antechrestos We also stick to boxed types (`Boolean` vs. `boolean`) because the documentation about what's guaranteed to be non-null is so poor.

PR #364, opened 2016-01-22
Get Service Plan

antechrestos:
This change adds the api to retrieve a specified service plan (```GET /v2/service_plans/:guid```)
See [this story](https://www.pivotaltracker.com/projects/816799/stories/101451578)


nebhale:
@antechrestos Notice the new naming scheme for ids.  We're getting away from unqualified `id` and moving to qualified `servicePlanId`, etc.

antechrestos:
@nebhale: agreed,  it will be clearer

PR #368, opened 2016-01-28
Delete Service Key

antechrestos:
This change adds the api to delete a service key (```DELETE /v2/service_keys/:guid```)
According to [this story](https://www.pivotaltracker.com/projects/816799/stories/101451554)


glyn:
Nice work @antechrestos, but I'll leave it to @nebhale to prioritise and merge.

PR #378, opened 2016-02-04
Fix loggregator get recent log

antechrestos:
Reading bytes of response asynchronously leads to being reading them while response input stream is closed

I implemented a good test that shows the issue (but did not find out how to integrate it with mock requests..
It fails without the fix, and success with the fix. If you want me to add a tests somewhere to complete the job, please show me how/where I can do this.

```java
@Test
public void recent() throws Throwable{
        final AtomicInteger count = new AtomicInteger();
        final CountDownLatch latch = new CountDownLatch(1);
        final AtomicReference<Throwable> error = new AtomicReference<>(null);
        Subscriber<LoggregatorMessage> subscriber = new Subscriber<LoggregatorMessage>(){
            @Override
            public void onSubscribe(Subscription s) {
                s.request(Long.MAX_VALUE);
            }

            @Override
            public void onNext(LoggregatorMessage message) {
                count.incrementAndGet();
            }

            @Override
            public void onError(Throwable t) {
                error.set(t);
                latch.countDown();
            }

            @Override
            public void onComplete() {
                latch.countDown();
            }
        };
        Publisher<LoggregatorMessage> publisher = loggregatorClient.recent(RecentLogsRequest.builder()
            .applicationId("application-id")
            .build());
        publisher.subscribe(subscriber);
        if (!latch.await(10, TimeUnit.SECONDS)) {
            throw new IllegalStateException("Subscriber timed out");
        }else if(error.get() != null){
            throw error.get();
        }
}
```

### DIFF
--- a/cloudfoundry-client-spring/src/main/java/org/cloudfoundry/client/spring/util/Multipart.java
+++ b/cloudfoundry-client-spring/src/main/java/org/cloudfoundry/client/spring/util/Multipart.java
@@ -22,6 +22,7 @@ import reactor.fn.Consumer;
 import reactor.fn.Function;
 import reactor.rx.Stream;
 
+import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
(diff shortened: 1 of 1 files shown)


nebhale:
Ugh.  That whole thing needs to be blown up and made non-blocking :frowning:

PR #379, opened 2016-02-05
Use a java.io.InputStream instead of java.io.File for application upload

mheath:
File is too narrow scoped of an abstraction for providing application bits to be uploaded to Cloud Foundry. Using an InputStream is much more generic and allows for uploading applications that live off of the local filesystem.


nebhale:
Good eye.

nebhale:
Turns out the implementation doesn't quite work.  Guessing that Spring doesn't convert an `InputStreamReasource` properly into a multipart.  I'll check it out tomorrow or Monday.

Issue #387, opened 2016-02-10
Loggregator broken... 


antechrestos:
Once again :smile: 
It aims to get recent logs from api endpoint leading to a ```404 NOT FOUND``` error.

nebhale:
@antechrestos Sorry about that.  I'm actually working now on doing some work to the client setups that will allow us to start properly integration testing the logging stuff as well.  Getting that tested is the next thing on my list.

antechrestos:
@nebhale no problem :)

PR #388, opened 2016-02-10
Fix loggregator endpoint

antechrestos:
Request to get the recent logs is made on the loggregator endpoint, not on the API
See #387


antechrestos:
Resolved by [59d1906](https://github.com/cloudfoundry/cf-java-client/commit/59d1906bb4a37a2c1ed39b6eb11d6524d2ebd796)

PR #389, opened 2016-02-11
Delete Service

antechrestos:
This change adds the api to delete a Service (```DELETE /v2/services/:guid```)
See [this story](https://www.pivotaltracker.com/projects/816799/stories/101452050)

### DIFF
--- a/cloudfoundry-client/src/main/java/org/cloudfoundry/client/v2/services/Services.java
+++ b/cloudfoundry-client/src/main/java/org/cloudfoundry/client/v2/services/Services.java
@@ -24,6 +24,14 @@ import reactor.core.publisher.Mono;
 public interface Services {
 
     /**
+     * Makes the <a href="http://apidocs.cloudfoundry.org/214/services/delete_a_particular_service.html">Delete the Service</a> request
+     *
+     * @param request the Delete Service request
+     * @return the response from the Delete Service request
+     */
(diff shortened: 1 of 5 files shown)


cloudfoundry-client/src/main/java/org/cloudfoundry/client/v2/services/Services.java
```diff
@@ -24,6 +24,14 @@
 public interface Services {
 
     /**
+     * Makes the <a href="http://apidocs.cloudfoundry.org/214/services/delete_a_particular_service.html">Delete the Service</a> request
+     *
+     * @param request the Delete Service request
+     * @return the response from the Delete Service request
+     */
+    Mono<Void> delete(DeleteServiceRequest request);
```
nebhale: @antechrestos You won't know this but we've identified that despite what the documentation says, any time an API allows an `async` argument, it returns a `JobEntity`-based response.  I've [opened an issue](https://github.com/cloudfoundry/cloud_controller_ng/issues/536) and make the fix in a following commit.

antechrestos:
@nebhale : ok this is just like the ```ServicePlan``` then. I saw it on yesterday but I was not sure if it applied in this case

nebhale:
Yeah.  We're using the existence of `async` as an indicator.

PR #390, opened 2016-02-12
List Service Plans

antechrestos:
This change adds the api to list service plans for a service (```GET /v2/services/:guid/service_plans```)
See [this story](https://www.pivotaltracker.com/projects/816799/stories/101452052)
@nebhale : I assumed [the documentation](http://apidocs.cloudfoundry.org/214/services/list_all_service_plans_for_the_service.html) was wrong about the ```service_guid``` filter parameter. Since the aim of the service is getting service plan for a given service, filtering by service seems to be incorrect...And it helped me because having a ```@Singular``` ```serviceIds``` would have be a mess with the ```serviceId``` attribute :smiley:


nebhale:
@antechrestos You were absolutely right on the `service_guid` filter.  We open issues against them whenever we see it, although there is a [technical issue](https://github.com/cloudfoundry/cloud_controller_ng/issues/496) as to why they are listed.

nebhale:
Ah!  They even beat us to that discrepancy.  You can see it in the [latest version](http://apidocs.cloudfoundry.org/230/services/list_all_service_plans_for_the_service.html) of the doc.

antechrestos:
@nebhale ok everything is fine then :smile_cat:

Issue #393, opened 2016-02-15
Binding a Service I nstance to a Route takes parameters


antechrestos:
The version of the API taken to implement this feature was deprecated and did not support parameters.
Now it does. See [here](http://apidocs.cloudfoundry.org/230/service_instances/binding_a_service_instance_to_a_route_%28experimental%29.html)

PR #394, opened 2016-02-15
Binding a Service Instance to a Route - Add parameters

antechrestos:
This change takes into account that Binding a Service Instance to a Route can  take arbitrary parameters See #393


nebhale:
We actually have to be careful about how we merge in API changes (not getting incompatibly ahead of the current release), but I believe that this will be fine to bring in directly.

antechrestos:
@nebhale thanks

PR #403, opened 2016-02-23
Create Identity zone

antechrestos:
This change adds the api to create an identity zone (POST /identity-zones)
See [this story](https://www.pivotaltracker.com/projects/816799/stories/114245163)

@nebhale : few remarks
- ```GetTokenKeyRequest``` should be under *src/main/lombok* instead of *src/main/java*
- To implement the ```CreateIdentityZoneRequest``` I took a look at [the model code](https://github.com/cloudfoundry/uaa/blob/master/model/src/main/java/org/cloudfoundry/identity/uaa/zone/IdentityZone.java) to get the mandatory fields
- Following the same approach I supposed the same entity was returned by the *UAA* [server](https://github.com/cloudfoundry/uaa/blob/master/server/src/main/java/org/cloudfoundry/identity/uaa/zone/IdentityZoneEndpoints.java)
- I implemented the test folowwing the same logic

As usual I will take any of your remarks. :smile:

### DIFF
--- a/cloudfoundry-client/src/main/lombok/org/cloudfoundry/uaa/identityzonemanagement/CreateIdentityZoneRequest.java
+++ b/cloudfoundry-client/src/main/lombok/org/cloudfoundry/uaa/identityzonemanagement/CreateIdentityZoneRequest.java
@@ -1,0 +1,108 @@
+/*
+ * Copyright 2013-2016 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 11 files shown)


cloudfoundry-client/src/main/lombok/org/cloudfoundry/uaa/identityzonemanagement/CreateIdentityZoneRequest.java
```diff
@@ -0,0 +1,108 @@
...
+    @Getter(onMethod = @__(@JsonProperty("description")))
+    private final String description;
+
+    /**
+     * The id of the identity zone. When not provided, an identifier will be generated
+     *
+     * @param identityZoneId the identity zone id
+     * @return the identity zone id
+     */
+    @Getter(onMethod = @__({@JsonProperty("id"), @JsonInclude(NON_NULL)}))
```
nebhale: In general we don't need the `@JsonInclude(NON_NULL)` annotation.  The `ObjectMapper` used by the `RestTemplate` is [configured for that](https://github.com/cloudfoundry/cf-java-client/blob/master/cloudfoundry-client-spring/src/main/lombok/org/cloudfoundry/spring/util/network/OAuth2RestTemplateBuilder.java#L121), for all properties.

PR #406, opened 2016-02-25
Update User Provided Service Instance

antechrestos:
This change adds the api to update a user provided service instance (```PUT /v2/user_provided_service_instances/:guid```)
See [this story](https://www.pivotaltracker.com/projects/816799/stories/101452068)

@nebhale : Few remarks
- I assumed [documentation](http://apidocs.cloudfoundry.org/231/user_provided_service_instances/updating_a_user_provided_service_instance.html) contained errors on two points: ```name``` is not mandatory and response status code is ```ACCEPTED``` instead of ```CREATED```
- I caught sight of the creation/update parameter ```route_service_url``` that leads to two response attributes ```route_service_url``` and ```route_url```. It is fixed in #407


nebhale:
@antechrestos The `name` being required is definitely wrong and I've opened [an issue](https://github.com/cloudfoundry/cloud_controller_ng/issues/545).  The return value is less interesting to us because we don't handle it, but it's possible that the value actually is `ACCEPTED`.  I'd guess that it should actually be something like `OK` (`ACCEPTED` is typically used to indicate something async, but that's a really rough guideline) but we don't care all that much.

PR #408, opened 2016-02-25
Get Identity Zone

antechrestos:
this change adds the api to get an identity zone (```GET /identity-zones/{identityZoneId}```)
See [this story](https://www.pivotaltracker.com/projects/816799/stories/114245231)
@nebhale : caught sight of that date were not JSON string but timestamp in ```IdentityZoneEntity```. Typed them as ```java.util.Date```

### DIFF
--- a/cloudfoundry-client-spring/src/test/java/org/cloudfoundry/spring/uaa/identityzonemanagement/SpringIdentityZoneManagementTest.java
+++ b/cloudfoundry-client-spring/src/test/java/org/cloudfoundry/spring/uaa/identityzonemanagement/SpringIdentityZoneManagementTest.java
@@ -19,10 +19,16 @@ package org.cloudfoundry.spring.uaa.identityzonemanagement;
 import org.cloudfoundry.spring.AbstractApiTest;
 import org.cloudfoundry.uaa.identityzonemanagement.CreateIdentityZoneRequest;
 import org.cloudfoundry.uaa.identityzonemanagement.CreateIdentityZoneResponse;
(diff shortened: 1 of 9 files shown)


cloudfoundry-client-spring/src/test/java/org/cloudfoundry/spring/uaa/identityzonemanagement/SpringIdentityZoneManagementTest.java
```diff
@@ -47,7 +53,7 @@ protected RequestContext getRequestContext() {
         @Override
         protected CreateIdentityZoneResponse getResponse() {
             return CreateIdentityZoneResponse.builder()
-                .createdAt("2015-07-27T22:43:20Z")
+                .createdAt(new Date(1426258488910L))
```
nebhale: The problem here is actually, that it shouldn't be date at all.  The type is a `Long` and while we know that it represents millis-since-epoch, the `-client` APIs are supposed to map as closely as possible to the original data types.  So the `CreateIdentityZoneResponse` needs to actually use `Long`s rather than `Date`s.

nebhale: I'll take care of the change in a following commit.

antechrestos: Just done it

PR #410, opened 2016-02-26
List Identity Zones

antechrestos:
This change adds the api to list identity zones (```GET /identity-zones```)
See [this story](https://www.pivotaltracker.com/projects/816799/stories/114245183)


nebhale:
@antechrestos We've updated to Java 8, so you can rebase and refactor if you'd like.  No obligation though; if you don't I'll take care of it on Monday.

antechrestos:
@nebhale good to read. Rebase done.

nebhale:
@antechrestos After thinking about it for a while, I've made the `ListIdentityZonesResponse` a `List<IdentityZone>` directly.  This is the correct logical mapping based on the payload so I want to stick close to that.  I'm not committed to this, so if we see it being a problem in the future, I'll take another look at the decision.

PR #411, opened 2016-02-29
Delete Identity Zone

antechrestos:
This change adds the api to delete an identity zone (```DELETE /identity-zones/{identityZoneId}```)
See [this story](https://www.pivotaltracker.com/projects/816799/stories/114245315)
@nebhale : strange, according to the api ```DELETE``` verb returns ```OK``` status **and** a the entity that will be removed...

### DIFF
--- a/cloudfoundry-client-spring/src/main/lombok/org/cloudfoundry/spring/uaa/identityzonemanagement/SpringIdentityZoneManagement.java
+++ b/cloudfoundry-client-spring/src/main/lombok/org/cloudfoundry/spring/uaa/identityzonemanagement/SpringIdentityZoneManagement.java
@@ -20,6 +20,8 @@ import lombok.ToString;
 import org.cloudfoundry.spring.util.AbstractSpringOperations;
 import org.cloudfoundry.uaa.identityzonemanagement.CreateIdentityZoneRequest;
 import org.cloudfoundry.uaa.identityzonemanagement.CreateIdentityZoneResponse;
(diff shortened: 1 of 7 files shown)


nebhale:
I get the feeling that this is a result of the fact that the Java  `Map.remove()` method returns the previous object if it existed.  Turns out nearly everyone lets their implementation leak through. :frowning:

nebhale:
@antechrestos https://cloudfoundry.slack.com/archives/uaa/p1456781925000113

cloudfoundry-client-spring/src/main/lombok/org/cloudfoundry/spring/uaa/identityzonemanagement/SpringIdentityZoneManagement.java
```diff
@@ -49,11 +51,16 @@ public SpringIdentityZoneManagement(RestOperations restOperations, URI root, Sch
     }
 
     @Override
-    public Mono<CreateIdentityZoneResponse> create(CreateIdentityZoneRequest request) {
+    public Mono<CreateIdentityZoneResponse> create(final CreateIdentityZoneRequest request) {
```
nebhale: As a heads up, our use of `final` in method arguments was a side-effect of the use of anonymous inner classes earlier.  Now with Java 8 we won't need these any more and I'm on a (long) mission to remove all of the unnecessary ones.

antechrestos:
@nebhale ok. Do you want me to remove them?

nebhale:
No, I'll take care of it as part of merging in.

PR #415, opened 2016-03-07
Create Identity Provider

antechrestos:
This change adds the api to create an identity provider (```POST /identity-providers```)
See [this story](https://www.pivotaltracker.com/projects/816799/stories/114246037)

@nebhale : Few remarks
- according to doc ```config``` is said to be required but looking at the example and the code [here](https://github.com/cloudfoundry/uaa/blob/master/model/src/main/java/org/cloudfoundry/identity/uaa/provider/IdentityProvider.java) we see that it accepts ```null``` config
- field ```identityZoneId``` is set by a header. Any idea of annotation to specify it?
- I named the interface ```IdentityProviders```. Maybe the interface ```IdentityZoneManagement``` should be renamed to  ```IdentityZones```  ? Otherwise I should rename ```IdentityProviderManagerment```. If so please let me know I will rework this story.

### DIFF
--- a/cloudfoundry-client-spring/src/main/java/org/cloudfoundry/spring/util/RequestHeaderBuilder.java
+++ b/cloudfoundry-client-spring/src/main/java/org/cloudfoundry/spring/util/RequestHeaderBuilder.java
@@ -1,0 +1,67 @@
+/*
+ * Copyright 2013-2016 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.cloudfoundry.spring.util;
+
+import lombok.AccessLevel;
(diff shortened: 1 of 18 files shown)


nebhale:
https://github.com/cloudfoundry/uaa/issues/331

nebhale:
@antechrestos As far as the `identityZoneId` issue, we don't have an implementation that does that yet and we need to add it.  However, this is a large enough issue that I'm going to hold these PRs.  We're working towards a release tomorrow and a change that large would be a bit risky.  In general, you might want to take a day or two off while we get this release out the door.

antechrestos:
@nebhale what ```identityZoneId``` issue?
No problem, I will do other things :smile:

antechrestos:
@nebhale nevermind I understood...I was tired after sport :sleeping:

antechrestos:
@nebhale : I initated a developement of such a thing:
- `RequestHeader`: annotation that allows to declare a request header for the request object
- `RequestHeaderBuilder`: class that reads request object and populates the current request
- `AbstractSpringOperations`: update of the function `postWithBody` to allow a `Consumer<HttpHeaders>` as last parameter; for now I only overloaded the `post` method so as not to impact too much code... Maybe at therm the `get`, `put` and `delete` method should/could support this mechanism? :weary: 
- `AbstractApiTest.RequestContext`: add the `requestHeader` method to check the presence of a specific header
- Use this last method in the `SpringIdentityProvidersTest.Create` `getRequestContext` override 

Indeed you're right, it impacts a lot of code :sob: ...
But it works :laughing:

nebhale:
@antechrestos We're finally ready to start on the UAA, but in the meantime, they've restructured their documentation so that it's much easier to understand.  Part of that change though means that none of this previous work will fit.  I'm going to close this out and we can start over with the better documentation and issues.

PR #416, opened 2016-03-07
Filter out more IntelliJ artifacts from CM

gregturn:


nebhale:
@gregturn Those artifacts are intentionally checked into source control for the use of the primary developers of the platform.  They're also a useful check while developing, that you're working against the same environment as everyone else.

PR #417, opened 2016-03-07
Create Identity Zone Client

antechrestos:
This change adds the api to create an identity zone client (```POST /identity-zones/{identityZoneId}/clients```)
See [this story](https://www.pivotaltracker.com/projects/816799/stories/115150309)


nebhale:
@antechrestos We're finally ready to start on the UAA, but in the meantime, they've restructured their documentation so that it's much easier to understand.  Part of that change though means that none of this previous work will fit.  I'm going to close this out and we can start over with the better documentation and issues.

antechrestos:
@nebhale yeah better close everything,  because for me, unlike Metallica,  the memory does not remain.....

Issue #351, opened 2016-01-13
updateApplicationUris not mapping route correctly 


sboyatsis:
Most likely a very unique scenario but we have a cloud organization with with the following domains available.

`sandbox.lmig.com`
`pdc.sandbox.lmig.com`

An application was trying to add a URI to an application with this URI:
`stevebspringmusic-pdc.sandbox.lmig.com`

When using the `updateApplicaitonUris` method specifying that URI the result is that the domain of `pdc.sandbox.lmig.com` is used with a host of `stevebspringmusic` .  The mapped route is incorrectly created as 

`stevebspringmusic.pdc.sandbox.lmig.com`

Desired result would be with the dash instead of .  (`stevebspringmusic-pdc.sandbox.lmig.com`)

In looking at cf client code it looks like the `extractUriInfo` function logic may need to be re-worked.  It determines that domain is `pdc.sandbox.lmig.com` and then grabs host name by subtracting 1 assuming that character is a "."  .  I've verified this behavior occurs in releases 1.1.2, 1.1.3, and 1.1.4.RELEASE.  I used this code segment to recreate the issue.

```
package com.lmig.cf.cfclienttest;
import org.cloudfoundry.client.lib.CloudCredentials;
import org.cloudfoundry.client.lib.CloudFoundryClient;
import org.cloudfoundry.client.lib.domain.CloudSpace;
import org.cloudfoundry.client.lib.CloudFoundryException;
import org.cloudfoundry.client.lib.CloudFoundryOperations;
import org.cloudfoundry.client.lib.StartingInfo;
import org.cloudfoundry.client.lib.domain.ApplicationStats;
import org.cloudfoundry.client.lib.domain.CloudApplication;
import org.cloudfoundry.client.lib.domain.CloudDomain;
import org.cloudfoundry.client.lib.domain.CloudInfo;
import org.cloudfoundry.client.lib.domain.CloudRoute;
import org.cloudfoundry.client.lib.domain.CloudService;
import org.cloudfoundry.client.lib.domain.CloudServiceOffering;
import org.cloudfoundry.client.lib.domain.Staging;

import java.util.ArrayList;
import java.util.Arrays;
import java.util.List;
import java.net.MalformedURLException;
import java.net.URI;
import java.net.URL;


/**
 * Testing misc cloudfoundry java client functions
 *
 */
public class App 
{
    public static void main( String[] args )
    {
    	String target = "https://api.sandbox.sys.paas.lmig.com" ;
[...]

PR #418, opened 2016-03-09
v1.1.2 Fix for issue #351

LokeshN:
Fixing the issue #351, for Host name containing the domain, check for the exact match of the domain.
 Checking for the exact match for the domain in the uris, by extracting the part from first "." in the uri.
Example, if the uri is "hostname-1.domain.com", taking the substring starting from first index of "." i.e, "domain.com" and comparing with the domain.

Added the integration test in CloudFoundryClientTest, test name: checkHostContainingDomain()


nebhale:
Closing as described in #351

PR #424, opened 2016-03-22
List Application Usage Events

antechrestos:
This change adds the api to list application usage events (`GET /v2/app_usage_events`)
See [this story](https://www.pivotaltracker.com/projects/816799/stories/101516688)

### DIFF
--- a/cloudfoundry-client/src/main/lombok/org/cloudfoundry/client/v2/appusageevents/ApplicationUsageEventEntity.java
+++ b/cloudfoundry-client/src/main/lombok/org/cloudfoundry/client/v2/appusageevents/ApplicationUsageEventEntity.java
@@ -1,0 +1,193 @@
+/*
+ * Copyright 2013-2016 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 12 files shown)


cloudfoundry-client/src/main/lombok/org/cloudfoundry/client/v2/appusageevents/ApplicationUsageEventEntity.java
```diff
@@ -0,0 +1,193 @@
...
+    private final String buildpackId;
+
+    /**
+     * The buildpack name
+     *
+     * @param buildpackName the build pack name
+     * @return the buildpack name
+     */
+    private final String buildpackName;
+
```
nebhale: The links in the issues don't help (they were made before we had the `latest-release` URI alias, but there are a couple of fields missing because you were looking at `214` instead of `232` the latest.  I'll add them in and just try to keep in mind that you want to look at the "latest" version of the documentation.

nebhale: Nope.  I'm absolutely wrong.  Apparently they just added some documentation for the metadata payload...  Sorry about the noise.

PR #427, opened 2016-03-24
Get Application Usage Event

antechrestos:
This change adds the api to get an application usage event (`GET /v2/app_usage_events/:guid`See [this story](https://www.pivotaltracker.com/projects/816799/stories/101516696)
@nebhale : I fixed a naming issue in `serviceusageevents`:

* GetServiceUsageEvent**s**Response => GetServiceUsageEventResponse
* GetServiceUsageEvent**s**Request => GetServiceUsageEventRequest 
* ServiceUsageEvent**s**Entity => GetServiceUsageEventEntity

I also fixed a bad url in the *ServiceInstances* api I spotted few days ago.

### DIFF
--- a/cloudfoundry-client/src/main/java/org/cloudfoundry/client/v2/serviceinstances/ServiceInstances.java
+++ b/cloudfoundry-client/src/main/java/org/cloudfoundry/client/v2/serviceinstances/ServiceInstances.java
@@ -24,8 +24,7 @@ import reactor.core.publisher.Mono;
 public interface ServiceInstances {
 
     /**
-     * Makes the <a href="http://apidocs.cloudfoundry.org/latest-release/service_instances/binding_a_service_instance_to_a_route_%28experimental%29.html">Bind Service Instance To a Route
-     * (experimental)</a> request
+     * Makes the <a href="http://apidocs.cloudfoundry.org/233/service_instances/binding_a_service_instance_to_a_route.html">Bind Service Instance To a Route</a> request
      *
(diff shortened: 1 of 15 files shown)


cloudfoundry-client/src/main/java/org/cloudfoundry/client/v2/serviceinstances/ServiceInstances.java
```diff
@@ -24,8 +24,7 @@
 public interface ServiceInstances {
 
     /**
-     * Makes the <a href="http://apidocs.cloudfoundry.org/latest-release/service_instances/binding_a_service_instance_to_a_route_%28experimental%29.html">Bind Service Instance To a Route
-     * (experimental)</a> request
+     * Makes the <a href="http://apidocs.cloudfoundry.org/233/service_instances/binding_a_service_instance_to_a_route.html">Bind Service Instance To a Route</a> request
```
nebhale: We're using the `latest-release` alias in these URIs these days.  I'll update.

antechrestos: @nebhale yeah sorry it was to point that it was not experimental any more

nebhale: No worries.  Any easy mistake to make and and easy fix.  This is why we do reviews.  You can focus on the hard stuff and I can make sure that the minor details are right :)

PR #434, opened 2016-03-29
Create Shared Domain - API

LokeshN:
This change adds the api to create a shared domain (POST /v2/shared_domains)


nebhale:
@LokeshN Thanks for such a high-quality submission.  I've made some changes that will follow in a polishing commit, but there are all formatting and code-style changes, nothing functional.  The code styling configuration checked into source control automatically enforces all of the changes I've made, so when you make another submission in the future, it'd be greatly appreciated if you could run the formatter before committing.

The bigger issue is that you've created a single PR with two separate units of work in it.  We intentionally separate each discrete API into its own commit so that we have a clear record of each one going in.  You've conflated [#101527354](https://www.pivotaltracker.com/story/show/101527354) and [#106155532](https://www.pivotaltracker.com/story/show/106155532) in this PR.  I'll be splitting that code out of this first commit and adding it back in as part of a second commit to follow.  Again, for the next time, please focus on a single API for each submission.

Again, thanks for such a high quality contribution.

LokeshN:
Thank you..

PR #439, opened 2016-04-01
Delete Route Mapping - API

LokeshN:
This change adds an api to delete the route mapping (DELETE /v2/route_mappings/routemapping-id)

I have done the formatting in the IDE, let me know if there are still issues related to formatting.

Thanks,
Lokesh

### DIFF
--- a/cloudfoundry-client-spring/src/test/java/org/cloudfoundry/spring/client/v2/routemappings/SpringRouteMappingsTest.java
+++ b/cloudfoundry-client-spring/src/test/java/org/cloudfoundry/spring/client/v2/routemappings/SpringRouteMappingsTest.java
@@ -19,12 +19,17 @@ package org.cloudfoundry.spring.client.v2.routemappings;
 import org.cloudfoundry.client.v2.Resource;
 import org.cloudfoundry.client.v2.routemappings.CreateRouteMappingRequest;
 import org.cloudfoundry.client.v2.routemappings.CreateRouteMappingResponse;
+import org.cloudfoundry.client.v2.routemappings.DeleteRouteMappingRequest;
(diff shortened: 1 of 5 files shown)


cloudfoundry-client-spring/src/test/java/org/cloudfoundry/spring/client/v2/routemappings/SpringRouteMappingsTest.java
```diff
@@ -79,4 +84,38 @@ protected CreateRouteMappingRequest getValidRequest() throws Exception {
         }
     }
 
+    public static final class Delete extends AbstractApiTest<DeleteRouteMappingRequest, DeleteRouteMappingResponse> {
```
nebhale: This is missing a test for `DeleteAsync`.  Since the payload of async operations is different (the `JobEntity` is returned for polling) we maintain a separate test for that as well.  I'll add it in a follow up commit.

cloudfoundry-client/src/main/lombok/org/cloudfoundry/client/v2/routemappings/DeleteRouteMappingResponse.java
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * Response object for the delete route mapping
+ */
+@Data
+@EqualsAndHashCode(callSuper = true)
+@ToString(callSuper = true)
+public final class DeleteRouteMappingResponse extends Resource<JobEntity> {
+
+    DeleteRouteMappingResponse(@JsonProperty("entity") JobEntity entity,
```
nebhale: Missing the `@Builder` annotation.  I think you would have caught this with the `DeleteAsync` test (you can't create the return type for that test without it), and I'll add this in a following commit.

PR #441, opened 2016-04-02
Domains List Operation

LokeshN:
This change adds the feature for domain list operations api


nebhale:
@LokeshN There are a couple of issues with this submission, but the biggest is that it doesn't list the same set of domains that the CLI does. If you run `CF_TRACE=true cf domains` you will see that it makes the following REST calls:

```
GET /v2/organizations/<ORG-ID>/private_domains HTTP/1.1
GET /v2/shared_domains HTTP/1.1
```

We would be attempting to replicate this behavior, so your submission will need to be updated to handle both requests.

I'll also make some other comments on the code itself with other areas for improvement.

cloudfoundry-operations/src/main/java/org/cloudfoundry/operations/domains/DefaultDomains.java
```diff
@@ -58,6 +61,22 @@ public DefaultDomains(CloudFoundryClient cloudFoundryClient) {
...
 
+    @Override
+    public Flux<Domain> list(ListDomainsRequest request) {
+        return ValidationUtils
+            .validate(request)
+            .flatMap(request1 -> getDomains(this.cloudFoundryClient, request1))
+            .map(resource -> toDomain(resource));
+    }
+
+    private static Flux<DomainResource> getDomains(CloudFoundryClient cloudFoundryClient, ListDomainsRequest request) {
```
nebhale: @LokeshN We isolate methods that make network calls, as you've done here, but by convention we'd call them what they were requesting.  In this case it would be `requestListDomains()`.  In a full implementation (see my comment on the main PR) we'd expect `requestListPrivateDomains()` and `requestListSharedDomains()`.

cloudfoundry-operations/src/main/lombok/org/cloudfoundry/operations/domains/Domain.java
```diff
@@ -0,0 +1,57 @@
...
+package org.cloudfoundry.operations.domains;
+
+import lombok.Builder;
+import lombok.Data;
+
+/**
+ * Domain object representation
+ */
+@Data
+public final class Domain {
```
nebhale: @LokeshN Based on the output for the CLI

```
$: cf domains
Getting domains in org bhale as bhale@pivotal.io...
name        status   
cfapps.io   shared   
```

we'd expect two fields `name` and `status`, plus a `domainId` to enable easy transfer back to the `-client` API.

cloudfoundry-operations/src/main/lombok/org/cloudfoundry/operations/domains/ListDomainsRequest.java
```diff
@@ -0,0 +1,68 @@
...
+import org.cloudfoundry.ValidationResult;
+import org.cloudfoundry.client.v2.InFilterParameter;
+import org.cloudfoundry.client.v2.PaginatedRequest;
+
+import java.util.List;
+
+@Data
+@EqualsAndHashCode(callSuper = true)
+@ToString(callSuper = true)
+public final class ListDomainsRequest extends PaginatedRequest implements Validatable {
```
nebhale: @LokeshN Based on the inputs to the CLI

```
$: cf help domains
NAME:
   domains - List domains in the target org

USAGE:
   cf domains
```

There is no need for a request object.  And example of this would be [`Organizations.list()` operation](https://github.com/cloudfoundry/cf-java-client/blob/master/cloudfoundry-operations/src/main/java/org/cloudfoundry/operations/organizations/Organizations.java#L56).

cloudfoundry-operations/src/test/java/org/cloudfoundry/operations/domains/DefaultDomainsTest.java
```diff
@@ -78,6 +84,38 @@ private static void requestOrganizationsEmpty(CloudFoundryClient cloudFoundryCli
                     .build()));
     }
 
+    private static void requestDomains(CloudFoundryClient cloudFoundryClient, String name, String ownerId) {
```
nebhale: @LokeshN When the mainline code is refactored with the two network requests, these methods should match the newly introduced method names in the main code.  This mapping between the two makes it easy for us to identify network requests that need to be mocked out.

cloudfoundry-operations/src/test/java/org/cloudfoundry/operations/domains/ListDomainsRequestTest.java
```diff
@@ -0,0 +1,68 @@
...
+import org.cloudfoundry.ValidationResult;
+import org.junit.Test;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import static org.cloudfoundry.ValidationResult.Status.VALID;
+import static org.junit.Assert.assertEquals;
+
+public final class ListDomainsRequestTest {
```
nebhale: No need for this test when the type goes away.

cloudfoundry-operations/src/test/java/org/cloudfoundry/operations/domains/DefaultDomainsTest.java
nebhale: @LokeshN These methods should also be sorted alphabetically.

LokeshN:
Thanks for the review Ben. I have updated the PR with the changes.

cloudfoundry-operations/src/main/lombok/org/cloudfoundry/operations/domains/Domain.java
```diff
@@ -0,0 +1,57 @@
...
+     * @param name the name returns the name of the domain
+     */
+    private final String domainName;
+
+    /**
+     * The status of the domain indicating shared or private domain
+     *
+     * @param status indicating the domain is shared or private
+     */
+    private final String status;
```
nebhale: @LokeshN Since this is a small, well-known list, we'd make an enumeration here.  I'll update.

cloudfoundry-operations/src/main/lombok/org/cloudfoundry/operations/domains/Domain.java
```diff
@@ -0,0 +1,57 @@
...
+ */
+@Data
+public final class Domain {
+
+    /**
+     * The GUID of the domain
+     *
+     * @param domainId the GUID of the domain return the GUID of the domain
+     */
+    private final String domainId;
```
nebhale: @LokeshN We've been inconsistent to date, but we're standardizing on simple names in the Operations layer, e.g. `id` and `name` instead of `domainId` and `domainName`.  I'll update.

PR #444, opened 2016-04-07
Add buildpacks operation

antechrestos:
This change brings the operation to list the buildpacks and tests for this changes.
See [this story](https://www.pivotaltracker.com/projects/816799/stories/106155552)

### DIFF
--- a/cloudfoundry-operations/src/main/java/org/cloudfoundry/operations/buildpacks/DefaultBuildpacks.java
+++ b/cloudfoundry-operations/src/main/java/org/cloudfoundry/operations/buildpacks/DefaultBuildpacks.java
@@ -1,0 +1,59 @@
+/*
+ * Copyright 2013-2016 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.cloudfoundry.operations.buildpacks;
+
+import org.cloudfoundry.client.CloudFoundryClient;
+import org.cloudfoundry.client.v2.buildpacks.BuildpackResource;
+import org.cloudfoundry.client.v2.buildpacks.ListBuildpacksRequest;
+import org.cloudfoundry.util.PaginationUtils;
+import org.cloudfoundry.util.ResourceUtils;
+import reactor.core.publisher.Flux;
+
+public final class DefaultBuildpacks implements Buildpacks {
+
+    private final CloudFoundryClient cloudFoundryClient;
+
+    public DefaultBuildpacks(CloudFoundryClient cloudFoundryClient) {
+        this.cloudFoundryClient = cloudFoundryClient;
+    }
+
+    @Override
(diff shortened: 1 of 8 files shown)


nebhale:
@antechrestos Big day; stepping up to the Operations API.  Let's see how you do :grinning:

cloudfoundry-operations/src/main/lombok/org/cloudfoundry/operations/buildpacks/BuildpackSummary.java
```diff
@@ -0,0 +1,92 @@
...
+package org.cloudfoundry.operations.buildpacks;
+
+import lombok.Builder;
+import lombok.Data;
+
+/**
+ * A Cloud Foundry Buildpack Summary
+ */
+@Data
+public final class BuildpackSummary {
```
nebhale: @antechrestos We've gone back and forth on this quite a bit, but our current behavior is that _at the operations layer_ we're going to use the simplest possible names (`Buildpack` instead of `BuildpackSummary`) **unless** two related operations return a `*Detail` / `*Summary` pair.

I'll make this change, and don't feel bad that you didn't get it right.  I've had the team working on a final decision and it's likely that our current code is inconsistent.

cloudfoundry-operations/src/main/java/org/cloudfoundry/operations/buildpacks/DefaultBuildpacks.java
```diff
@@ -0,0 +1,59 @@
...
+        return PaginationUtils
+            .requestResources(page -> cloudFoundryClient.buildpacks()
+                .list(ListBuildpacksRequest.builder()
+                    .page(page)
+                    .build()));
+    }
+
+    private static BuildpackSummary toBuildpackResource(BuildpackResource resource) {
+        return BuildpackSummary.builder()
+            .enabled(ResourceUtils.getEntity(resource).getEnabled())
```
nebhale: @antechrestos Extract the `ResourceUtils.getEntity(resource)` into a local variable.  I'll update.

cloudfoundry-operations/src/main/java/org/cloudfoundry/operations/buildpacks/DefaultBuildpacks.java
```diff
@@ -0,0 +1,59 @@
...
+package org.cloudfoundry.operations.buildpacks;
+
+import org.cloudfoundry.client.CloudFoundryClient;
+import org.cloudfoundry.client.v2.buildpacks.BuildpackResource;
+import org.cloudfoundry.client.v2.buildpacks.ListBuildpacksRequest;
+import org.cloudfoundry.util.PaginationUtils;
+import org.cloudfoundry.util.ResourceUtils;
+import reactor.core.publisher.Flux;
+
+public final class DefaultBuildpacks implements Buildpacks {
```
nebhale: @antechrestos Beautiful work.  Dead on following our design and implementation standards.

cloudfoundry-operations/src/test/java/org/cloudfoundry/operations/buildpacks/DefaultBuildpacksTest.java
```diff
@@ -0,0 +1,68 @@
...
+
+        @Before
+        public void setUp() throws Exception {
+            requestBuildpacks(this.cloudFoundryClient);
+        }
+
+        @Override
+        protected void assertions(TestSubscriber<BuildpackSummary> testSubscriber) {
+            testSubscriber
+                .assertEquals(fill(BuildpackSummary.builder(), "buildpack-")
```
nebhale: @antechrestos Again, likely inconsistent, but I like the contents of the test itself (both the value in `invoke()` and `assertions()`) to be complete.  `fill()` certainly gives the correct result, but I prefer to see the inputs and outputs directly in the test code rather than having to go look up (or remember) how what the generated values would be.  I'll update.

PR #445, opened 2016-04-07
Create User Provided Service Operation

antechrestos:
This change adds the operation to create a user provided service instance
It also create an abstract class of User Provided Service Instance Resource and makes all response class inherit from it
@nebhale : I willingly added the `routeServiceUrl`as it is now a parameter in the [documentations] (http://apidocs.cloudfoundry.org/latest-release/user_provided_service_instances/creating_a_user_provided_service_instance.html)
See [story](https://www.pivotaltracker.com/projects/816799/stories/106155500)

### DIFF
--- a/cloudfoundry-client/src/main/lombok/org/cloudfoundry/client/v2/userprovidedserviceinstances/AbstractUserProvidedServiceInstanceResource.java
+++ b/cloudfoundry-client/src/main/lombok/org/cloudfoundry/client/v2/userprovidedserviceinstances/AbstractUserProvidedServiceInstanceResource.java
@@ -16,24 +16,22 @@
 
 package org.cloudfoundry.client.v2.userprovidedserviceinstances;
 
-
 import com.fasterxml.jackson.annotation.JsonProperty;
-import lombok.Builder;
 import lombok.Data;
 import lombok.EqualsAndHashCode;
 import lombok.ToString;
+import org.cloudfoundry.client.v2.Resource;
 
 /**
- * The resource response payload for the Update User Provided Service Instance Response
+ * The resource response payload for User Provided Service Instances
  */
 @Data
 @EqualsAndHashCode(callSuper = true)
 @ToString(callSuper = true)
-public final class UpdateUserProvidedServiceInstanceResponse extends AbstractUserProvidedServiceInstanceResource {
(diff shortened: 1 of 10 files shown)


cloudfoundry-client/src/main/lombok/org/cloudfoundry/client/v2/userprovidedserviceinstances/AbstractUserProvidedServiceInstanceResource.java
```diff
@@ -0,0 +1,38 @@
...
+import lombok.ToString;
+import org.cloudfoundry.client.v2.Resource;
+
+/**
+ * The resource response payload for User Provided Service Instances
+ */
+@Data
+@EqualsAndHashCode(callSuper = true)
+@ToString(callSuper = true)
+public abstract class AbstractUserProvidedServiceInstanceResource extends Resource<UserProvidedServiceInstanceEntity> {
```
nebhale: @antechrestos Solid work.  This is a refactoring that @Zteve has been doing, so it's great to see you picking up on it.  It's the perfect thing to do here.

cloudfoundry-operations/src/main/java/org/cloudfoundry/operations/services/DefaultServices.java
```diff
@@ -184,6 +194,12 @@ private static ServiceInstanceType convertToInstanceType(String type) {
             .cast(AbstractServiceInstanceResource.class);
     }
 
+    private static Mono<AbstractUserProvidedServiceInstanceResource> createUserProvidedServiceInstance(CloudFoundryClient cloudFoundryClient, String spaceId,
+                                                                                                       CreateUserProvidedServiceInstanceRequest request) {
+        return requestCreateUserProvidedServiceInstance(cloudFoundryClient, spaceId, request.getName(), request.getCredentials(), request.getRouteServiceUrl(), request.getSyslogDrainUrl())
+            .cast(AbstractUserProvidedServiceInstanceResource.class);
```
nebhale: @antechrestos This can be simpler (no need for the `.cast()`) and inlined into the main flow.  I'll update.

nebhale:
@antechrestos Great work. For future reference, the examples in the operations descriptions can be out of date, just like the client APIs.  Running the same commands (`cf help create-user-provided-service` in this case) will return what we expect to be implemented.

PR #447, opened 2016-04-12
use HttpComponentsClientHttpRequestFactory with HttpClient

robertgmoss:
for connections instead of CustomSslSimpleClientHttpRequestFactory


nebhale:
@robertgmoss We're actually removing the entire Spring-based network stack this week, to be replaced by a Netty-based one.  I'll leave this PR open until that is done to remind me that I need to test against HTTP endpoints, but I don't foresee this being included.

nebhale:
@robertgmoss What's the use case for accessing HTTP endpoints instead of HTTPS?  We've finished the network stack migration, but it does assume HTTPS only.  Before I modify that, I want to get an idea why you wouldn't always used HTTPS.

robertgmoss:
I don't recall the details, but using the HTTP endpoint fixed an issue we had connecting to our client's CF wherby an SSLException was thrown - an issue that the cli didn't have.  Agree in general HTTPS is better, though.

nebhale:
@robertgmoss We've found some issues here and there in months since this was opened.  Can you have a crack with the latest snapshot and see if an HTTPS-only solution now works for you?

nebhale:
Since there's been no response on this issue in a couple of weeks, I'm going to close it.  If you'd like to see it re-opened, please comment on the issue and I'll reopen it.

PR #449, opened 2016-04-12
Quotas Operation

antechrestos:
This changes adds the operation to retrieve the quotas
See [here](https://www.pivotaltracker.com/n/projects/816799/stories/106155582)
@nebhale I based my code on this [source](https://github.com/cloudfoundry/cli/blob/master/cf/commands/quota/quotas.go)

### DIFF
--- a/cloudfoundry-operations/src/main/java/org/cloudfoundry/operations/quotas/DefaultQuotas.java
+++ b/cloudfoundry-operations/src/main/java/org/cloudfoundry/operations/quotas/DefaultQuotas.java
@@ -1,0 +1,62 @@
+/*
+ * Copyright 2013-2016 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.cloudfoundry.operations.quotas;
+
+import org.cloudfoundry.client.CloudFoundryClient;
+import org.cloudfoundry.client.v2.organizationquotadefinitions.ListOrganizationQuotaDefinitionsRequest;
+import org.cloudfoundry.client.v2.organizationquotadefinitions.OrganizationQuotaDefinitionEntity;
+import org.cloudfoundry.client.v2.organizationquotadefinitions.OrganizationQuotaDefinitionResource;
+import org.cloudfoundry.util.PaginationUtils;
+import org.cloudfoundry.util.ResourceUtils;
+import reactor.core.publisher.Flux;
+
+public final class DefaultQuotas implements Quotas {
+
+    private final CloudFoundryClient cloudFoundryClient;
+
+    public DefaultQuotas(CloudFoundryClient cloudFoundryClient) {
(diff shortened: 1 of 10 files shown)


nebhale:
This is going to be another big one (naming ambiguity), but I'll leave some notes in the commit for next time.  The biggest change is that the name of the package and type is tied to CLI heading:

```
USER ADMIN:
   create-user                            Create a new user
   delete-user                            Delete a user

   org-users                              Show org users by role
   set-org-role                           Assign an org role to a user
   unset-org-role                         Remove an org role from a user

   space-users                            Show space users by role
   set-space-role                         Assign a space role to a user
   unset-space-role                       Remove a space role from a user

ORG ADMIN:
   quotas                                 List available usage quotas
   quota                                  Show quota info
   set-quota                              Assign a quota to an org

   create-quota                           Define a new resource quota
   delete-quota                           Delete a quota
   update-quota                           Update an existing resource quota

   share-private-domain                   Share a private domain with an org
   unshare-private-domain                 Unshare a private domain with an org

SPACE ADMIN:
   space-quotas                           List available space resource quotas
   space-quota                            Show space quota info
   create-space-quota                     Define a new space resource quota
   update-space-quota                     Update an existing space quota
   delete-space-quota                     Delete a space quota definition and unassign the space quota from all spaces
   set-space-quota                        Assign a space quota definition to a space
   unset-space-quota                      Unassign a quota from a space
````

That value is then reflected in a tag on the PT issue as well.  Renaming is easy for me to do, so don't worry to much about it, just a heads up for next time.

File: cloudfoundry-operations/src/main/lombok/org/cloudfoundry/operations/quotas/Quota.java [hunk 1]
```diff
@@ -0,0 +1,101 @@
...
+package org.cloudfoundry.operations.quotas;
+
+import lombok.Builder;
+import lombok.Data;
+
+/**
+ * A Cloud Foundry Quota
+ */
+@Data
+public final class Quota {
```
nebhale: @antechrestos We've decided to add an `id` to each Operations type so that it's possible to drop down to the Client API when necessary.  I'll add.

cloudfoundry-operations/src/main/lombok/org/cloudfoundry/operations/quotas/Quota.java
```diff
@@ -0,0 +1,101 @@
...
+     */
+    private final Boolean allowPaidServicePlans;
+
+    /**
+     * The application instance limit
+     *
+     * @param applicationInstanceLimit the application instance limit
+     * @return applicationInstanceLimit
+     */
+    private final Integer applicationInstanceLimit;
```
nebhale: @antechrestos This one is interesting.  It's not in the CLI output, but is probably worth having.  Wonder if that constitutes a bug in the CLI.

cloudfoundry-operations/src/main/java/org/cloudfoundry/operations/quotas/DefaultQuotas.java
```diff
@@ -0,0 +1,62 @@
...
+        this.cloudFoundryClient = cloudFoundryClient;
+    }
+
+    @Override
+    public Flux<Quota> list() {
+        return requestQuotas(this.cloudFoundryClient)
+            .map(DefaultQuotas::toQuotaResource);
+    }
+
+    private static Flux<OrganizationQuotaDefinitionResource> requestQuotas(CloudFoundryClient cloudFoundryClient) {
```
nebhale: @antechrestos In a class this small, this name is nice and unambiguous, but in the larger classes (see `DefaultApplications`) it'd rapidly be too vague.  The name should closer match the request actually made, in this case `requestListOrganizationQuotas`.  I'll update this one and the matching test method.

antechrestos:
@nebhale ok I will rework #451 according to your notes. Thanks!

PR #451, opened 2016-04-13
Marketplace operation

antechrestos:
This change adds the marketplace operation
See [this story](https://www.pivotaltracker.com/projects/816799/stories/106155472)

### DIFF
--- a/cloudfoundry-operations/src/main/java/org/cloudfoundry/operations/services/Services.java
+++ b/cloudfoundry-operations/src/main/java/org/cloudfoundry/operations/services/Services.java
@@ -64,6 +64,13 @@ public interface Services {
     Flux<ServiceInstance> listInstances();
 
     /**
+     * LList available offerings in the marketplace
+     *
+     * @return the service instances
+     */
+    Flux<ServiceOffering> listOfferings(GetMarketplaceRequest request);
+
+    /**
      * Unbind a service instance from an application
      *
      * @param request the unbind service instance request
--- a/cloudfoundry-operations/src/main/lombok/org/cloudfoundry/operations/services/ServiceOffering.java
+++ b/cloudfoundry-operations/src/main/lombok/org/cloudfoundry/operations/services/ServiceOffering.java
@@ -1,0 +1,74 @@
+/*
+ * Copyright 2013-2016 the original author or authors.
+ *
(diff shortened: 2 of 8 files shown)


antechrestos:
@nebhale reworked according to #449 and the [commit](https://github.com/cloudfoundry/cf-java-client/commit/c0342f6e0a6db9bd180becd51be84f0ac4f41709) you made to refactor

cloudfoundry-operations/src/main/java/org/cloudfoundry/operations/services/Services.java
```diff
@@ -64,6 +64,13 @@
     Flux<ServiceInstance> listInstances();
 
     /**
+     * LList available offerings in the marketplace
+     *
+     * @return the service instances
+     */
+    Flux<ServiceOffering> listOfferings(GetMarketplaceRequest request);
```
nebhale: @antechrestos I'd like to unify this signature and types to be consistently "marketplace".  I think `ServiceOffering` is fine though.  I'll update.

nebhale: @antechrestos Nope, I'm wrong.  The change needs to be that the request type is `ListServiceOfferingsRequest`.  The rest if fine.  I'll still make the change.

cloudfoundry-operations/src/main/lombok/org/cloudfoundry/operations/services/ServiceOffering.java
```diff
@@ -0,0 +1,74 @@
...
+     */
+    private final String label;
+
+    /**
+     * The plan names
+     *
+     * @param planNames the plan names
+     * @return the plan names
+     */
+    private final List<String> planNames;
```
nebhale: @antechrestos This doesn't seem to expose the free/paid and plan description metadata.  I'm thinking that rather than a `String` here, it needs to be a richer data type.  Thoughts?

antechrestos:
@nebhale I am up to take a plane. I will take a look tomorrow

antechrestos:
@nebhale and rename other things

nebhale:
@antechrestos I'm having a go now.  No need for you to take a look.

PR #452, opened 2016-04-13
List Running Security Groups

antechrestos:
This change adds the api to list running security groups (`GET /v2/config/running_security_groups`)
See [this story](https://www.pivotaltracker.com/projects/816799/stories/101522638)


nebhale:
@antechrestos This one is a bit weird.  You pretty much followed the rules to a T, but the documentation for it is very confusing.  I believe that this should be part of the "security groups" grouping.  I'll make the updates.

PR #453, opened 2016-04-14
Create buildpack operation

LokeshN:
The change adds the operation for create-builpack. 

In the REST client, the buildpack creation api needs only the name and the remaining arguments are optional but in CF cli, name, filename and position arguments are mandatory for calling the create-buildpack, so I have done the additional validation check in the create operation method.

Let me know whether this is fine or we can add the mandatory arguments check in the CreateBuildRequest itself.

### DIFF
--- a/cloudfoundry-operations/src/main/java/org/cloudfoundry/operations/buildpacks/Buildpacks.java
+++ b/cloudfoundry-operations/src/main/java/org/cloudfoundry/operations/buildpacks/Buildpacks.java
@@ -16,12 +16,22 @@
 
 package org.cloudfoundry.operations.buildpacks;
 
+import org.cloudfoundry.client.v2.buildpacks.CreateBuildpackRequest;
 import reactor.core.publisher.Flux;
+import reactor.core.publisher.Mono;
 
 /**
  * Main entry point to the Cloud Foundry Buildpacks Operations API
  */
 public interface Buildpacks {
+
+    /**
+     * Create a new Buildpack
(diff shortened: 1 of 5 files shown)


nebhale:
@LokeshN I think what you've missed is that even in the case where the type names are the same and the behaviors are similar, we actually create an operations-specific type.  So I would expect _another_ `CreateBuildpackRequest` that implements `Validatable` as input to the operation.  The contents of that type are then translated to the Client API invocation.  Validation is operations-specific as well.  So take a look at a simple implementation of [another operation](https://github.com/cloudfoundry/cf-java-client/blob/master/cloudfoundry-operations/src/main/java/org/cloudfoundry/operations/organizationadmin/DefaultOrganizationAdmin.java#L40-L45) and you'll see the pattern.

LokeshN:
@nebhale I have updated the PR with the changes
Also for the create buildpack operation, the buildpack is enabled by default, so I am setting the enable property in request to true, if it is null.

nebhale:
Nope, accidental closure.

LokeshN:
@nebhale does this PR needs some more changes..

nebhale:
@LokeshN It does.  One of the things that it does not do is actually upload the bits of the buildpack itself.  I'm doing acceptance of the client API to enable this now and then I was going to let you know.

LokeshN:
@nebhale Will create a separate PR for this one as I am not able to merge latest changes with my branch.

PR #455, opened 2016-04-14
Get Organization Quota operation

antechrestos:
This change adds the operation to get an organization quota by its name
See [this story](https://www.pivotaltracker.com/projects/816799/stories/106155584)


cloudfoundry-operations/src/main/java/org/cloudfoundry/operations/organizationadmin/DefaultOrganizationAdmin.java
```diff
@@ -33,15 +39,30 @@ public DefaultOrganizationAdmin(CloudFoundryClient cloudFoundryClient) {
...
+    public Mono<OrganizationQuota> getQuota(GetQuotaRequest getQuotaRequest) {
+        return ValidationUtils.validate(getQuotaRequest)
+            .then(request -> getOrganizationQuota(this.cloudFoundryClient, request.getName()))
+            .map(DefaultOrganizationAdmin::toOrganizationQuota);
+    }
+
+    @Override
     public Flux<OrganizationQuota> listQuotas() {
-        return requestListOrganizationQuotas(this.cloudFoundryClient)
+        return requestListOrganizationQuotas(this.cloudFoundryClient, Function.identity())
```
nebhale: @antechrestos In general, instead of adding these filtering functions we typically just have individual request methods for various input argument collections.  It makes it a bit easier to read and reason about.

PR #456, opened 2016-04-17
Set Organization Quota Operation

antechrestos:
This change adds the operation to set an organization quota
See [here](https://www.pivotaltracker.com/projects/816799/stories/106155586)


nebhale:
Resolved by ffc97265459613b5a527166ca0bd2f72926d043a

PR #457, opened 2016-04-18
Rename Service Operation

antechrestos:
This change adds the operation to rename a service
See [this story](https://www.pivotaltracker.com/projects/816799/stories/106155486)
@nebhale Executing the command of the *cli* with `CF_TRACE=true` shows that it requests the `/v2/services/:guid` with the the `id`got from service plan but does not seem to do anything with it as it only ends with a `PUT` on `/v2/service_instances/:guid`. Hence, I willingly skipped this part. Tell me if I missed anything.

### DIFF
--- a/cloudfoundry-operations/src/main/java/org/cloudfoundry/operations/services/DefaultServices.java
+++ b/cloudfoundry-operations/src/main/java/org/cloudfoundry/operations/services/DefaultServices.java
@@ -30,11 +30,14 @@ import org.cloudfoundry.client.v2.servicebindings.DeleteServiceBindingResponse;
 import org.cloudfoundry.client.v2.servicebindings.ListServiceBindingsRequest;
 import org.cloudfoundry.client.v2.servicebindings.ServiceBindingResource;
 import org.cloudfoundry.client.v2.serviceinstances.AbstractServiceInstanceResource;
+import org.cloudfoundry.client.v2.serviceinstances.BaseServiceInstanceEntity;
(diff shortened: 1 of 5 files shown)


nebhale:
@antechrestos It seems that it isn't quite complete.  In the case of a managed service, the following happens:

```
GET /v2/spaces/<ID>/service_instances?return_user_provided_service_instances=true&q=name%3A<NAME>&inline-relations-depth=1 HTTP/1.1
GET /v2/services/<ID>
PUT /v2/service_instances/<ID>?accepts_incomplete=true
```

I agree that the middle request is unnecessary.

In the case of a user-provided service, the following happens.
```
GET /v2/spaces/<ID>/service_instances?return_user_provided_service_instances=true&q=name%3A<NAME>&inline-relations-depth=1
PUT /v2/user_provided_service_instances/<ID>
```

So, dispatching on the type in `resources.entity.type` is necessary.

antechrestos:
@nebhale Ok I will change it

antechrestos:
@nebhale I implemented it. I am not fully satisfied with my implementation. The only solution I found to detect if it is a user provided service instance is by looking at [the code](https://github.com/cloudfoundry/cli/blob/master/cf/models/service_instance.go)
Is there a function in reactor such as Mono.if().else()?

nebhale:
@antechrestos I'm going to do a polishing commit, bit it's mostly splitting up methods and some renaming.  Fundamentally your if/else structure is correct.  It's a fine line knowing when imperative is better than reactive, but this reads very clearly, so it's right.

PR #458, opened 2016-04-19
Set Running Security Group

antechrestos:
This change adds the api to set a running security group as default for Running Applications (`PUT /v2/config/running_security_groups/:guid`)
See [this story](https://www.pivotaltracker.com/projects/816799/stories/101522640)
@nebhale what do you think of renaming the functions:
1. `deleteRunningDefault` :arrow_right: `deleteRunningSecurityGroup`
2. `listRunningDefaults` :arrow_right:  `listRunningSecurityGroups`
Therefore the `setRunningDefault` may be named `setRunningSecurityGroupAsDefault`


Zteve:
Hi Benjamin,
My own opinion is that we should avoid the use of the term 'delete' altogether and use these:
`setRunningDefault`, `setStagingDefault`, `unsetRunningDefault`, `unsetStagingDefault`, `listRunningDefaults` and `listStagingDefaults`. The latter list methods are, in fact, unnecessary at the operations layer (and have problems of their own anyway—see https://github.com/cloudfoundry/cloud_controller_ng/issues/578) but we include them for completeness.

We would normally avoid putting the root term `SecurityGroup` in the method name: this is the name of the class from which the method comes (well, its plural is), and we don't do that anywhere else.

The `DELETE` requests don't actually delete anything.

antechrestos:
@Zteve Oki doki.

nebhale:
@Zteve Do we have another example of the set/unset pairing in the code today?

Zteve:
The closest thing we have to set/unset are things like associate; but I'm not aware of a DELETE being used to take an element out of a collection (or to set a `boolean` in it to `false`, which is all this seems to do). The collections involved are not 'addressable' (they don't have an id) either.  Strange. But par for the course, I feel.

PR #467, opened 2016-04-22
Set Security Group Staging Default

antechrestos:
This change adds the api to set a security group staging as default (`PUT /v2/config/staging_security_groups/:guid`)
See [here](https://www.pivotaltracker.com/projects/816799/stories/101522652)


nebhale:
resolved by 83115d2e8409ce6674dbcc5cc4ccf2361fbc1b3d

PR #470, opened 2016-04-25
Delete Service Instance Operation

antechrestos:
This change adds the operation to delete a service instance


nebhale:
@antechrestos Missing the functionality to delete the user provided service instance.  I'll let you add that in and resubmit.

antechrestos:
@nebhale and now?

PR #471, opened 2016-04-25
Create buildpack Operation

LokeshN:
The change adds the operation for create-builpack.

### DIFF
--- a/cloudfoundry-operations/src/main/java/org/cloudfoundry/operations/buildpacks/DefaultBuildpacks.java
+++ b/cloudfoundry-operations/src/main/java/org/cloudfoundry/operations/buildpacks/DefaultBuildpacks.java
@@ -19,10 +19,8 @@ package org.cloudfoundry.operations.buildpacks;
 import org.cloudfoundry.client.CloudFoundryClient;
 import org.cloudfoundry.client.v2.buildpacks.BuildpackEntity;
 import org.cloudfoundry.client.v2.buildpacks.BuildpackResource;
-import org.cloudfoundry.client.v2.buildpacks.CreateBuildpackRequest;
 import org.cloudfoundry.client.v2.buildpacks.CreateBuildpackResponse;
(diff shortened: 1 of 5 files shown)


LokeshN:
@nebhale I have updated the PR with uploading of bits , let me know if this is ok.

cloudfoundry-operations/src/main/java/org/cloudfoundry/operations/buildpacks/DefaultBuildpacks.java
```diff
@@ -33,11 +42,34 @@ public DefaultBuildpacks(CloudFoundryClient cloudFoundryClient) {
...
             .map(DefaultBuildpacks::toBuildpackResource);
     }
 
+    private static Mono<CreateBuildpackResponse> requestCreateBuildpack(CloudFoundryClient cloudFoundryClient, String buildpackName, Integer position, Boolean enable) {
+        return cloudFoundryClient.buildpacks()
+            .create(org.cloudfoundry.client.v2.buildpacks.CreateBuildpackRequest
+                .builder()
+                .name(buildpackName)
+                .position(position)
+                .enabled(enable)
```
nebhale: @LokeshN Rather than encapsulating the "default" behavior in the request, we choose to apply it where it is used.  So in this case, we'd do an `Optional.ofNullable(enable).orElse(true)`.  I'll make the change.

cloudfoundry-operations/src/main/lombok/org/cloudfoundry/operations/buildpacks/CreateBuildpackRequest.java
```diff
@@ -0,0 +1,104 @@
...
+@Data
+public final class CreateBuildpackRequest implements Validatable {
+
+    /**
+     * The buildpack name
+     *
+     * @param name the buildpack name
+     * @return the buildpack name
+     */
+    private final String name;
```
nebhale: @LokeshN Alphabetical order.  I'll update.

PR #476, opened 2016-05-04
List service keys operation

LokeshN:
This change adds the operation for listing the service keys for an
service instance


nebhale:
@LokeshN I'm on these.  Give me a day to clarify something with the CAPI team and I'll merge them.

PR #481, opened 2016-05-13
List Service Brokers Operation

LokeshN:
This change adds the List Service Brokers operation

### DIFF
--- a/cloudfoundry-operations/src/main/java/org/cloudfoundry/operations/serviceadmin/ServiceAdmin.java
+++ b/cloudfoundry-operations/src/main/java/org/cloudfoundry/operations/serviceadmin/ServiceAdmin.java
@@ -1,0 +1,33 @@
+/*
+ * Copyright 2013-2016 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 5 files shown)


cloudfoundry-operations/src/main/lombok/org/cloudfoundry/operations/serviceadmin/ServiceBroker.java
```diff
@@ -0,0 +1,50 @@
...
+package org.cloudfoundry.operations.serviceadmin;
+
+import lombok.Builder;
+import lombok.Data;
+
+/**
+ * A Service Broker
+ */
+@Data
+public final class ServiceBroker {
```
nebhale: @LokeshN Missing an `id` parameter.  While this doesn't strictly match the CLI UI, we've made the decision that every artifact returned from the `Operations` API should have a matching `-client` id to facility dropping down to the lower level API.  I'll add this in a following commit.

cloudfoundry-operations/src/main/java/org/cloudfoundry/operations/serviceadmin/ServiceAdmin.java
```diff
@@ -0,0 +1,33 @@
...
+
+package org.cloudfoundry.operations.serviceadmin;
+
+
+import reactor.core.publisher.Flux;
+
+/**
+ * Main entry point to the Cloud Foundry Service Admin API
+ */
+public interface ServiceAdmin {
```
nebhale: @LokeshN This interface is never exposed from the `CloudFoundryOperations` API.  I'll add this in a following commit.

PR #486, opened 2016-05-28
Upgrade identity zones

antechrestos:
This change adds the support for configuration in five operations 
It also makes `description` `@Nullable` in an update context as it is [optional](http://docs.cloudfoundry.com/uaa/#updating-an-identity-zone)


cloudfoundry-client/src/main/java/org/cloudfoundry/uaa/identityzones/_IdentityZoneConfiguration.java
```diff
@@ -0,0 +1,135 @@
...
+
+    /**
+     * The token policy
+     */
+    @JsonDeserialize(using = TokenPolicyDeserializer.class)
+    @JsonProperty("tokenPolicy")
+    @Nullable
+    abstract TokenPolicy getTokenPolicy();
+
+    static final class LinksDeserializer extends StdDeserializer<Links> {
```
nebhale: @antechrestos You don't need to add custom deserializers for these as Jackson will do the right thing.  The custom deserializer pattern should only be necessary for [things with polymorphic return types](https://github.com/cloudfoundry/cf-java-client/blob/master/cloudfoundry-client/src/main/java/org/cloudfoundry/client/v3/packages/Package.java#L92-L99).

PR #487, opened 2016-05-29
Update User

antechrestos:
This change adds the apmi to update a user (`PUT /Users/:user-id`)
@nebhale Still a lack when dealing with  [specifics headers](http://docs.cloudfoundry.com/uaa/#update49):
I got catch of `IdentityZoned` interface but doing the same thing may lead to creating a new interface as I did (`IfMatched`), annotating the abstract method in abstract request class  with the `@JsonIgnore`annotation, and sadest part, either implement a *builder* that handle such class as it is done [here](https://github.com/cloudfoundry/cf-java-client/blob/master/cloudfoundry-client-spring/src/main/java/org/cloudfoundry/reactor/uaa/IdentityZoneBuilder.java)
 In the current implementation I just did like this:

```java
       if (request instanceof IdentityZoned) {
            IdentityZoned identityZoned = (IdentityZoned) request;
            outbound.addHeader("X-Identity-Zone-Id", identityZoned.getIdentityZoneId());
        }
        if (request instanceof IfMatched) {
            IfMatched ifMatched = (IfMatched) request;
            outbound.addHeader("If-Match", ifMatched.getIfMatch());
        }
```

But that implies to rename the builder.

Maybe create an annotation and a builder that scans the class?

### DIFF
--- a/cloudfoundry-client-spring/src/main/java/org/cloudfoundry/reactor/uaa/AbstractUaaOperations.java
+++ b/cloudfoundry-client-spring/src/main/java/org/cloudfoundry/reactor/uaa/AbstractUaaOperations.java
@@ -40,6 +40,14 @@ public abstract class AbstractUaaOperations extends AbstractReactorOperations {
         super(authorizationProvider, httpClient, objectMapper, root);
     }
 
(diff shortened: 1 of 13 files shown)


antechrestos:
@nebhale follwing your advice I added an other `builder` and call it from `AbstractUaaOperation`

PR #493, opened 2016-06-04
Delete Service Broker

LokeshN:
This change adds the delete service broker operation


nebhale:
@LokeshN Really high quality stuff.  Thanks for the contribution.

PR #497, opened 2016-06-08
Identity providers

antechrestos:
This change brings the full api for identity providers
- Create: `POST /identity-providers`
- Update: `PUT /identity-providers/:id`
- Retrieve : `GET /identity-providers/:id`
- List: `GET /identity-providers`

Bonus track: the other way to implement `data` field in `lifecycle` object


cloudfoundry-client/src/main/java/org/cloudfoundry/client/v3/_Lifecycle.java
```diff
@@ -35,7 +32,11 @@
     /**
      * The datas
      */
-    @JsonDeserialize(using = DataDeserializer.class)
+    @JsonTypeInfo(use = JsonTypeInfo.Id.NAME, include = JsonTypeInfo.As.EXTERNAL_PROPERTY, property = "type")
```
nebhale: @antechrestos I desperately want this to work, but I'm getting compilation errors with it.  Any chance you can double check compiling this with clean, from the command line?

nebhale:
@antechrestos Yeah, I'm almost sure that there's a bug in Immutables around this.  If you agree, can you please [open an issue against the project](https://github.com/immutables/immutables/issues) and CC me on it?

PR #523, opened 2016-07-15
Create Client

antechrestos:
This change brings the api to create a client (POST `/oauth/clients`)
@twoseat @nebhale I spotted a little issue in the [documentation](http://docs.cloudfoundry.com/uaa/#create76): in the example, the `autoapprove`, supposed to be an `Array` of `String` (I checked their code), is set with a `boolean` value...

### DIFF
--- a/cloudfoundry-client/src/test/java/org/cloudfoundry/uaa/clients/CreateClientRequestTest.java
+++ b/cloudfoundry-client/src/test/java/org/cloudfoundry/uaa/clients/CreateClientRequestTest.java
@@ -16,27 +16,21 @@
 
 package org.cloudfoundry.uaa.clients;
 
-import reactor.core.publisher.Mono;
+import org.junit.Test;
 
-/**
- * Main entry point to the UAA Clients API
- */
-public interface Clients {
+public final class CreateClientRequestTest {
 
-    /**
-     * Makes the <a href="">Create Client</a> request
-     *
-     * @param request Create Client request
-     * @return the Response to the Create Client Request
-     */
-    Mono<CreateClientResponse> create(CreateClientRequest request);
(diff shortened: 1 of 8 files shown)


cloudfoundry-client/src/test/java/org/cloudfoundry/uaa/clients/CreateClientRequestTest.java
```diff
@@ -0,0 +1,36 @@
...
+    @Test(expected = IllegalStateException.class)
+    public void noId() {
+        CreateClientRequest.builder()
+            .build();
+    }
+
+    @Test
+    public void valid() {
+        CreateClientRequest.builder()
+            .clientId("test-client-id")
```
Zteve: Are they all optional except  this one?

antechrestos:
@zteve the only required is this one. The problem you are pointing is that the other one are collections and the check is not done as the collection itself is not null... Yet empty. Maybe we may need some annotation `@NotEmpty`...

Zteve:
Yes, I see.  This is a problem which we will have to address separately (it already exists in `UpdateClientRequest` and is similar to a problem we had in `UpdateApplicationRequest`).

In this case, although we don't mandate the parameter be set on the `Request`, and "default" it to empty; this is not what the user expects or the interface defines.  Omitting this parameter should be an error at `build()` time of the request.  (If we mark this as `@Nullable` and omit it it doesn't appear in the payload, which is a real error.)  It is perfectly reasonable for the user to set this parameter to an empty list, so we shouldn't mandate `@NonEmpty`, but this should not be a "default".

I have a solution to this, and will address this in a different story, and this PR/commit can stand. Thank you.

PR #525, opened 2016-07-18
Fix add member implementation

antechrestos:
The current implementation of the `AddMemberRequest` did not illustrate the fact that all parameters are required
See [the documentation](http://docs.cloudfoundry.com/uaa/#add-member)


twoseat:
@antechrestos I've merged this in (but apparently forgot the magic setting that would have closed this PR). Technically it wasn't needed - recent changes made to `AbstractMember` meant that the fields were actually required. But we've also decided to avoid using the same abstract class for both requests and responses (the fact that the request and the response might look the same is effectively a coincidence, rather than being a necessity) so I've included your change for that reason. Thanks!

PR #526, opened 2016-07-19
Fix Create/Update client requests

antechrestos:
Set all list attributes as `@Nullable` in order not to send an empty collection as default value.
Add a check on required attribute.


Zteve:
Thanks Ben!  I made some minor changes (as usual). See the story https://www.pivotaltracker.com/story/show/126488735 for details.

PR #528, opened 2016-07-20
Add tags in service offering class

alexander071:
Currently, the controller returns a list of tags for each service offering
exposed in the /services endpoint. These tags, however, are ignored by the
Java client and are not part of the ServiceOffering class that is returned
to the user.

These tags can be useful to tools that automate the management of services
(to determine whether a service's tags should be updated  or not). Because
of this, this commit adds a getTags( ) method in the ServiceOffering class
that allows users to access them.


Zteve:
Hi Alexander,

In general the operations API in `cf-java-client` is intended to mimic the function provided by the CF CLI. For this reason the information output by the CLI is normally only what appears in the responses from the operations API calls.  The one exception to that is that we often put object (primary) ids in the responses. For example, the service offering ids are included in the response you are looking at.

Full information is available in the client API layer by using the `services.get(GetServiceRequest)` on the `CloudFoundryClient` interface, which uses this id. This includes the tags you are looking for.

When more detailed information and manipulation is required, as appears to be in your case, it is likely that you would need to use the lower-level client API (which is intended to be complete). If there is anything missing in this layer, then we would gladly entertain a bug report.

Rather than migrate all the information from the client API into the operations layer, I recommend use of the client API when needed.

With this understanding, I'm going to close this PR.

PR #531, opened 2016-07-21
List clients

antechrestos:
This change adds the API to list clients (`GET /oauth/clients`)


Zteve:
Another one spot on.  Well done.

PR #532, opened 2016-07-21
Authorized grant type

antechrestos:
As the values for client attribute authorized grant types are restricted to specific OAuth 2 process grant type values, this field can be implemented with an enum value


Zteve:
Indeed they can.

PR #542, opened 2016-08-01
Batch Delete Clients

antechrestos:
This change brings the api to delete clients in batch (`POST /oauth/clients/tx/delete`)

### DIFF
--- a/cloudfoundry-client-reactor/src/test/java/org/cloudfoundry/reactor/uaa/clients/ReactorClientsTest.java
+++ b/cloudfoundry-client-reactor/src/test/java/org/cloudfoundry/reactor/uaa/clients/ReactorClientsTest.java
@@ -21,6 +21,8 @@ import org.cloudfoundry.reactor.TestRequest;
 import org.cloudfoundry.reactor.TestResponse;
 import org.cloudfoundry.reactor.uaa.AbstractUaaApiTest;
 import org.cloudfoundry.uaa.SortOrder;
+import org.cloudfoundry.uaa.clients.BatchDeleteClientsRequest;
+import org.cloudfoundry.uaa.clients.BatchDeleteClientsResponse;
 import org.cloudfoundry.uaa.clients.Client;
 import org.cloudfoundry.uaa.clients.CreateClientRequest;
 import org.cloudfoundry.uaa.clients.CreateClientResponse;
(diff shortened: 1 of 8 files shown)


cloudfoundry-client-reactor/src/test/java/org/cloudfoundry/reactor/uaa/clients/ReactorClientsTest.java
```diff
@@ -53,6 +56,76 @@
...
+                    .scope("clients.read", "clients.write")
+                    .tokenSalt("48TIsq")
+                    .build())
+                .build();
+        }
+
+        @Override
+        protected BatchDeleteClientsRequest getValidRequest() throws Exception {
+            return BatchDeleteClientsRequest.builder()
+                .clientIdentification(ClientIdentification.builder()
```
Zteve: I'm not sure I like this interface—it is wordy and a little cumbersome for the user.  Also there is no prospect of the `ClientIdentification` class being used anywhere else.  I'll work on a way of simplifying this.

cloudfoundry-client/src/main/java/org/cloudfoundry/uaa/clients/_BatchDeleteClientsResponse.java
```diff
@@ -0,0 +1,64 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.cloudfoundry.uaa.clients;
+
+/**
```
Zteve: I'll remove this spurious comment.

cloudfoundry-client/src/test/java/org/cloudfoundry/uaa/clients/BatchDeleteClientsRequestTest.java
```diff
@@ -0,0 +1,38 @@
...
+    public void noClientIdentification() {
+        BatchDeleteClientsRequest.builder()
+            .build();
+    }
+
+    @Test
+    public void valid() {
+        BatchDeleteClientsRequest.builder()
+            .clientIdentification(ClientIdentification.builder()
+                .clientId("client-id")
```
Zteve: This should be "test-client-id", and is simpler anyway when I eliminate `ClientIdentification`.

antechrestos:
@zteve I will change it tomorrow. Please let me do it, I was not satisfied myself

antechrestos:
@zteve tomorrow morning gmt+2

Zteve:
OK; that's fine—I'd rather you did it anyway. I've had a happy time trying it myself, and we can see if you get there in the same way I did. :-)

antechrestos:
@Zteve : I did it the brutal way. I tried to look for a magic annotation but failed :disappointed:

Zteve:
@antechrestos : Ben; your code and mine were almost identical. So it must be excellent. :smile:

I made some very small adjustments (naming and javaDoc comment stuff) and the only thing worth mentioning is the place in the serializer where you construct a `List` of `Map`s to get it to produce the `[`...`]` JSon. I changed this to generate an _array_ of `Map`s instead (using `toArray()`), which is easier and simpler and produces exactly the same JSon anyway.

(Upon reflection, this isn't worth mentioning after all.)

Thank you for another piece of the work well done.

antechrestos:
@Zteve Yeah I did not know the `toArray` stream closure. Now I will :smirk:

PR #557, opened 2016-08-22
Polishing: replace duplicate code by util static function

Zteve:
Mono<byte[]> collectByteArray(Flux<byte[]> bytes) is pulled out to a static
method in ResourceUtils so that it is available to the four places it is used
(twice in integration tests and twice in Reactor unit tests). This actually
increases the runtime jar (util is available in production) but not the
runtime footprint. It is a reasonably useful method, however, and reduces the
interdependencies in the test code.


nebhale:
@Zteve The `collectByteArray()` method should probably be moved to `OperationsUtils` rather than `ResourceUtils` as it's an additional operation rather than something having to do with `org.cloudfoundry.client.v2.Resource`s.

Zteve:
Yup; agreed.

Issue #569, opened 2016-09-28
CF client Invalid auth token error 


mohanpindyala:
getting following error after a while

```
reactor.core.Exceptions$BubblingException: org.cloudfoundry.client.v2.CloudFoundryException: CF-InvalidAuthToken(1000): Invalid Auth Token
    at reactor.core.Exceptions.bubble(Exceptions.java:96)
    at reactor.core.publisher.Operators.onErrorDropped(Operators.java:317)
    at reactor.core.publisher.LambdaSubscriber.onError(LambdaSubscriber.java:126)
    at reactor.core.publisher.FluxPeek$PeekSubscriber.onError(FluxPeek.java:193)
    at reactor.core.publisher.MonoFlatMap$FlattenSubscriber.onError(MonoFlatMap.java:166)
    at reactor.core.publisher.MonoCollectList$MonoBufferAllSubscriber.onError(MonoCollectList.java:100)
    at reactor.core.publisher.FluxMap$MapSubscriber.onError(FluxMap.java:134)
```

any idea

using these versions for java client

```
<cloudfoundry-client-spring.version>2.0.0.RELEASE</cloudfoundry-client-spring.version>
<cloudfoundry-operations.version>2.0.0.RELEASE</cloudfoundry-operations.version>
<reactor-core.version>3.0.0.RELEASE</reactor-core.version>
<reactor-netty.version>0.5.0.RELEASE</reactor-netty.version>
```

twoseat:
@mohanpindyala I suspect this is because your token has expired. Is 'a while' the same period every time?

nebhale:
It does certainly appear that the token being presented to the API is invalid.  What does your `TokenProvider` configuration look like?

mohanpindyala:
there you go 
` CloudFoundryClient cfClient = ReactorCloudFoundryClient.builder()
                                                               .connectionContext(DefaultConnectionContext.builder()
                                                                                                          .apiHost(cfApiUrl)
                                                                                                          .skipSslValidation(skipSSLValidation)
                                                                                                          .build())
                                                               .tokenProvider(PasswordGrantTokenProvider.builder()
                                                                                                        .password
                                                                                                                (password)
                                                                                                        .username
                                                                                                                (username)
                                                                                                        .build())
                                                               .build();

        cfOperations = DefaultCloudFoundryOperations.builder()
                                                    .cloudFoundryClient(cfClient)
                                                    .organization(cforg)
                                                    .space(cfspace)
[...]

PR #576, opened 2016-11-10
Refresh token expiration

s-bortolussi:
If both access token and refresh token have expired, previous implementation fails to get a new access token.
This fix forces to get a new access token without using expired refresh token.

[#131326071][#569]


nebhale:
@s-bortolussi The latest snapshot should solve all problems around token negotiation and refresh.  If you could give the snapshot a try and report back with any issues, it'd be greatly appreciated.

Issue #551, opened 2016-08-09
Recent log: Some messages are not well read


antechrestos:
Working with the new version of `getRecentLog` lead to some not handled messages:
These errors are not systematic and are not always the same( sometime expecting VARINT and complaining about  not the good length..)

Here is some of these

```
java.lang.IllegalStateException: Unexpected call to beginMessage()
    at com.squareup.wire.ProtoReader.beginMessage(ProtoReader.java:90) ~[wire-runtime-2.2.0.jar:na]
    at org.cloudfoundry.dropsonde.events.ValueMetric$ProtoAdapter_ValueMetric.decode(ValueMetric.java:189) ~[cloudfoundry-client-2.0.0.M10.jar:2.0.0.M10]
    at org.cloudfoundry.dropsonde.events.ValueMetric$ProtoAdapter_ValueMetric.decode(ValueMetric.java:165) ~[cloudfoundry-client-2.0.0.M10.jar:2.0.0.M10]
    at org.cloudfoundry.dropsonde.events.Envelope$ProtoAdapter_Envelope.decode(Envelope.java:573) ~[cloudfoundry-client-2.0.0.M10.jar:2.0.0.M10]
    at org.cloudfoundry.dropsonde.events.Envelope$ProtoAdapter_Envelope.decode(Envelope.java:499) ~[cloudfoundry-client-2.0.0.M10.jar:2.0.0.M10]
    at com.squareup.wire.ProtoAdapter.decode(ProtoAdapter.java:195) ~[wire-runtime-2.2.0.jar:na]
    at com.squareup.wire.ProtoAdapter.decode(ProtoAdapter.java:189) ~[wire-runtime-2.2.0.jar:na]
    at org.cloudfoundry.reactor.doppler.ReactorDopplerEndpoints.toEnvelope(ReactorDopplerEndpoints.java:66) ~[cloudfoundry-client-reactor-2.0.0.M10.jar:2.0.0.M10]
    at org.cloudfoundry.reactor.doppler.ReactorDopplerEndpoints$$Lambda$38/751608431.apply(Unknown Source) ~[na:na]
    at reactor.core.publisher.FluxMap$MapConditionalSubscriber.onNext(FluxMap.java:220) ~[reactor-core-3.0.0.RC1.jar:na]
[...]

PR #582, opened 2016-11-25
Clean up of deprecated HttpStart and HttpStop entries

gberche-orange:
Update following doppler update: no longer returns HttpStart and HttpStop messages

Since [cf-release v242](https://github.com/cloudfoundry/cf-release/releases/tag/v242), the doppler endpoints no longer return HttpStart and HttpStop messages, in particular in the [apps/appid/stream](https://github.com/cloudfoundry/loggregator/blob/develop/src%2Ftrafficcontroller%2FREADME.md#endpoints) endpoint. 

> loggregator is No longer supporting message aggregation of HttpStart and HttpStop messages in Metron Agent.

This PR cleans up the deprecated protocol buffer structure and enveloppe objects associated. I would expect the wire lib to ignore these fields in previous versions of doppler protocol (prior to v242). This is part of an attempt to get a fix for ``recentLogs`` endpoint #551 (even though only LogMessage events are actually returned in this case)

----
More background:

See announcement at http://cf-dev.70369.x6.nabble.com/cf-dev-Proposal-to-deprecate-the-HttpStart-and-HttpStop-events-tt5507.html

Related dropsonde update https://github.com/cloudfoundry/dropsonde-protocol/commit/3dd1473786cf6acdcc55e2206a58f08ad02f29ca
and loggregator backlog entry https://www.pivotaltracker.com/n/projects/993188/stories/128893349


nebhale:
@gberche-orange I'm pretty sure that we can't remove these types for compatibility reasons.  Why do you think that removing the types would solve #551?

gberche-orange:
@nebhale 

> I'm pretty sure that we can't remove these types for compatibility reasons. 

I would think that a newer version of cf-java-client with this PR invoked against an older version of cf (prior to v242) returning HttpStart and HttpStop events would just ignore these fields that have absolute positions in the protocol buffer definition.

I understand this is similar to the cf CLI using the NOA library that indeed bumped the doppler protocol buffer definitions, and AFAIK did not break compatibility

>  Why do you think that removing the types would solve #551?

Sorry I was ambiguous here. It would not solve the recentLogs issue. I just came across this potential clean up when reproducing the issue 551. Indeed PR 582 is unrelated to issue 551.

nebhale:
@gberche-orange My compatibility argument is that I'd like to allow a wide-range of Java Client-to-CF compatibility.  The API might handle unknown payloads correctly, but if users are pointing at CF versions that still produce those messages legitimately, I think we should allow them through.  For better or for worse, we actually implement a _bunch_ of deprecated functionality all of the CAPI API as well.

gberche-orange:
thanks for the merge Ben.

Issue #479, opened 2016-05-09
UnknownHostException when behind a proxy


michahirsch:
I'm using the cloudfoundry client behind an proxy and I get problems since the change to the reactor Netty HttpClient. The Netty HttpClient is using a `io.netty.resolver.DefaultNameResolver` to resolve the hostname. The hostname should be resolved by the proxy which actually is working but using the `InetAddress.getByName(inetHost)` will always through an `UnknownHostException`.

Is there anyway to inject an own implementation of a `io.netty.resolver.InetNameResolver`?

```
Caused by: java.net.UnknownHostException: xxxxxxxxxxxxx
	at java.net.InetAddress.getAllByName0(InetAddress.java:1259)
	at java.net.InetAddress.getAllByName(InetAddress.java:1171)
	at java.net.InetAddress.getAllByName(InetAddress.java:1105)
	at java.net.InetAddress.getByName(InetAddress.java:1055)
	at io.netty.resolver.DefaultNameResolver.doResolve(DefaultNameResolver.java:40)
	at io.netty.resolver.SimpleNameResolver.resolve(SimpleNameResolver.java:62)
	at io.netty.resolver.SimpleNameResolver.resolve(SimpleNameResolver.java:53)
	at io.netty.resolver.InetSocketAddressResolver.doResolve(InetSocketAddressResolver.java:55)
	at io.netty.resolver.InetSocketAddressResolver.doResolve(InetSocketAddressResolver.java:31)
	at io.netty.resolver.AbstractAddressResolver.resolve(AbstractAddressResolver.java:106)
	at io.netty.bootstrap.Bootstrap.doResolveAndConnect(Bootstrap.java:174)
	at io.netty.bootstrap.Bootstrap.connect(Bootstrap.java:142)
```

nebhale:
@michahirsch We've never really had full proxy support and lucked out with the support that Spring's `RestTemplate` provided.  It's definitely on the roadmap, but not currently at the top.  I'll have a look today if there is something that we can do easily to get you back up and running.

nebhale:
[...]

Issue #602, opened 2017-01-03
DefaultConnectionContext.builder().skipSslValidation(true) no longer appears to work?


mchitwood:
I upgraded from 2.0.3 to 2.1.0 and receive a null exception when I try to create a `ReactorCloudFoundryClient`.

nebhale:
Can you please include the stack of the error?  Our testing disables SSL validation in many of our environments, so we believe that it works as it always has.

mchitwood:
java.nio.channels.UnresolvedAddressException
	at sun.nio.ch.Net.checkAddress(Unknown Source)
	at sun.nio.ch.SocketChannelImpl.connect(Unknown Source)
	at io.netty.channel.socket.nio.NioSocketChannel.doConnect(NioSocketChannel.java:331)
	at io.netty.channel.nio.AbstractNioChannel$AbstractNioUnsafe.connect(AbstractNioChannel.java:254)
	at io.netty.channel.DefaultChannelPipeline$HeadContext.connect(DefaultChannelPipeline.java:1266)
	at io.netty.channel.AbstractChannelHandlerContext.invokeConnect(AbstractChannelHandlerContext.java:556)
	at io.netty.channel.AbstractChannelHandlerContext.connect(AbstractChannelHandlerContext.java:541)
	at io.netty.channel.AbstractChannelHandlerContext.connect(AbstractChannelHandlerContext.java:523)
	at io.netty.channel.DefaultChannelPipeline.connect(DefaultChannelPipeline.java:985)
	at io.netty.channel.AbstractChannel.connect(AbstractChannel.java:255)
	at io.netty.bootstrap.Bootstrap$3.run(Bootstrap.java:252)
	at io.netty.util.concurrent.AbstractEventExecutor.safeExecute(AbstractEventExecutor.java:163)
	at io.netty.util.concurrent.SingleThreadEventExecutor.runAllTasks(SingleThreadEventExecutor.java:418)
	at io.netty.channel.nio.NioEventLoop.run(NioEventLoop.java:454)
	at io.netty.util.concurrent.SingleThreadEventExecutor$5.run(SingleThreadEventExecutor.java:873)
	at java.lang.Thread.run(Unknown Source)

nebhale:
I've just tried to connect to the address listed there and cannot connect to it:

```
wlan-10-81-130-97 master: telnet api.system.asv-pr.ice.predix.io 443
Trying 65.216.161.18...

...
```

[...]

PR #614, opened 2017-01-16
Configure Netty client with a resolved proxy host address

tareqhs:
Netty uses unresolved address in proxy mode due to proxy delegation. The address of the proxy host itself should be resolved before being set in the client options to avoid UnknownHostException failures, during connections context SSL initialization as well.

Fixes issue https://github.com/cloudfoundry/cf-java-client/issues/479

Signed-off-by: Tareq Sharafy <tareq.sha@gmail.com>


nebhale:
This is an issue in `reactor-netty` that we worked through, [some time ago](https://github.com/cloudfoundry/cf-java-client/issues/479) and has been [recently regressed](https://github.com/cloudfoundry/cf-java-client/issues/602#issuecomment-270381999).  The Reactor team knows about it and is working on fixing it for the next release.

nebhale:
We've isolated this issue and it's in the `reactor-netty` project.  It will require a change to that library which won't make the `2.4.0.RELEASE`, but as soon as it's available we'll do another release with it.

nebhale:
The latest snapshot `2.5.0.BUILD-SNAPSHOT` should fix this issue.  If you can find the time, please take a look at it to confirm that we've solved your problem.  Note that `reactor-core` should be `3.0.5.RELEASE` and `reactor-netty` should be `0.6.1.RELEASE` with this snapshot.

Issue #761, opened 2017-06-01
ChangeUserPasswordRequest builder insists on 'oldPassword' being set, but it really is optional and needed for a 'reset password' operation.


james-fitzgerald:
The REST API for changing a users password can also be used to reset the password if the user has appropriate credentials (i.e. `uaa.admin` + `password.write`). However the `ChangeUserPasswordRequest` builder insisted on the `oldPassword` value being specified and that needs to be omitted when resetting another users password.

nebhale:
@james-fitzgerald The [documentation](https://docs.cloudfoundry.org/api/uaa/#change-user-password) states it as `required`.  If you believe this is incorrect, can you please open an issue against the UAA and link it here?  Once they agree, we'll do the work to change it.

PR #762, opened 2017-06-01
The Rest api for changing a users password can also be used to reset …

james-fitzgerald:
…the password if the user has appropriate credentials (i.e. uaa.admin + password.write). However the ChangeUserPasswordRequest builder insisted on the 'oldPassword' value being specified and that needs to be omitted when resetting another users password.

This pull request addresses defect: https://github.com/cloudfoundry/cf-java-client/issues/761


james-fitzgerald:
This has already been fixed by nebhale https://github.com/cloudfoundry/cf-java-client/commit/f292ca601ff0d998d6be839f1d66bfb4a9e1ed90

PR #793, opened 2017-08-09
Use asMemoryInteger when parsing disk_quota from manifest files.

### LINKED ISSUE #792: Pushing manifests fails when `disk_quota` is specified with units
According to [Deploying with Application Manifests](https://docs.cloudfoundry.org/devguide/deploy-apps/manifest.html#disk-quota),

> This attribute requires a unit of measurement: `M`, `MB`, `G`, or `GB`, in upper case or lower case.

However when `disk_quota` is specified in this manner in the manifest file (for example: `disk_quota: 1024M`), `org.cloudfoundry.operations.applications.Applications.pushManifest()` fails with the following error:

```
org.cloudfoundry.client.v2.ClientV2Exception: CF-AppInvalid(100001): The app is invalid: disk_quota too little disk requested (must be greater than zero)
	at org.cloudfoundry.reactor.util.ErrorPayloadMapper.lambda$null$0(ErrorPayloadMapper.java:47)
	at org.cloudfoundry.reactor.util.ErrorPayloadMapper.lambda$null$10(ErrorPayloadMapper.java:108)
	at reactor.core.publisher.MonoThenMap$ThenMapMain.onNext(MonoThenMap.java:120)
        ...
```

I've stepped through the code in a debugger and I can see that when the manifest is parsed by `org.cloudfoundry.operations.applications.ApplicationManifestUtils.read()`, it sets `disk` to 0 instead of the value in the manifest file. Looks like it's caused by calling `asInteger` instead of `asMemoryInteger` when parsing the `disk_quota` field in [ApplicationManifestUtils.toApplicationManifest](https://github.com/cloudfoundry/cf-java-client/blob/v2.16.0.RELEASE/cloudfoundry-operations/src/main/java/org/cloudfoundry/operations/applications/ApplicationManifestUtils.java#L241).

I'll open a PR shortly

This issue was originally reported by @sergiofm in [JENKINS-46049](https://issues.jenkins-ci.org/browse/JENKINS-46049).

sworisbreathing:
Handle lowercase units and two-character units (i.e. 'mb'/'gb') in asMemoryInteger.
Fixes #792.

### DIFF
--- a/cloudfoundry-operations/src/main/java/org/cloudfoundry/operations/applications/ApplicationManifestUtils.java
+++ b/cloudfoundry-operations/src/main/java/org/cloudfoundry/operations/applications/ApplicationManifestUtils.java
@@ -163,10 +163,14 @@ public final class ApplicationManifestUtils {
             } else if (raw.isTextual()) {
                 String text = raw.asText();
 
-                if (text.endsWith("G")) {
+                if (text.toUpperCase().endsWith("G")) {
                     return Integer.parseInt(text.substring(0, text.length() - 1)) * GIBI;
-                } else if (text.endsWith("M")) {
+                } else if (text.toUpperCase().endsWith("GB")) {
+                    return Integer.parseInt(text.substring(0, text.length() - 2)) * GIBI;
+                } else if (text.toUpperCase().endsWith("M")) {
                     return Integer.parseInt(text.substring(0, text.length() - 1));
+                } else if (text.toUpperCase().endsWith("MB")) {
+                    return Integer.parseInt(text.substring(0, text.length() - 2));
                 } else {
                     return 0;
                 }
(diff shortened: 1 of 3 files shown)


sworisbreathing:
ok I've just emailed my signed CLA

sworisbreathing:
Closing to reopen as per the email I got about the ICLA

cloudfoundry-operations/src/main/java/org/cloudfoundry/operations/applications/ApplicationManifestUtils.java
```diff
@@ -163,10 +163,14 @@ private static void asMemoryInteger(JsonNode payload, String key, Consumer<Integ
             } else if (raw.isTextual()) {
                 String text = raw.asText();
```
Zteve: If you put the `toUpperCase()` here, it might read a bit better and not affect any of the parsing.

I'm suggesting:
```
 String text = raw.asText().toUpperCase();
```

and then:
```
               if (text.endsWith("G")) {
                      return Integer.parseInt(text.substring(0, text.length() - 1)) * GIBI;
               } else if (text.endsWith("GB")) {
                      return Integer.parseInt(text.substring(0, text.length() - 2)) * GIBI;
               } else if (text.endsWith("M")) {
                      return Integer.parseInt(text.substring(0, text.length() - 1));
               } else if (text.endsWith("MB")) {
                      return Integer.parseInt(text.substring(0, text.length() - 2));
               } else {
```
later on.

sworisbreathing: hrm... makes sense.

PR #821, opened 2017-10-10
Add unit test for failing manifest inheritance case

shakuzen:
There are existing tests for manifest inheritance and common properties, but there is an assumption that the application will be named in the base manifest. This is not documented[1] and is not an assumption held by the CF CLI - the test case added works with the CF CLI. This will need to be fixed to allow the Cloud Foundry Jenkins plugin[2] to easily support this use case.

[1] https://docs.cloudfoundry.org/devguide/deploy-apps/manifest.html#multi-manifests
[2] https://issues.jenkins-ci.org/browse/JENKINS-47271

---

When specifying only common properties in the base manifest, there will be no applications present, which causes the following logic to skip adding any of the common properties from the base manifest.

https://github.com/cloudfoundry/cf-java-client/blob/2f9c7eb99da6026545de49c3c67c07eb348c6305/cloudfoundry-operations/src/main/java/org/cloudfoundry/operations/applications/ApplicationManifestUtils.java#L200-L210

### DIFF
--- a/cloudfoundry-operations/src/test/resources/fixtures/base-manifest.yml
+++ b/cloudfoundry-operations/src/test/resources/fixtures/base-manifest.yml
@@ -1,0 +1,9 @@
+---
+buildpack: https://github.com/cloudfoundry/java-buildpack#v4.5.2
+memory: 1G
+path: /target/demo.jar
+env:
+  JBP_CONFIG_OPEN_JDK_JRE: '{ jre: { version: 1.8.0_+ } }'
+
+#applications:
+#- name: demo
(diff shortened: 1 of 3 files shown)


cloudfoundry-operations/src/test/resources/fixtures/base-manifest.yml
```diff
@@ -0,0 +1,9 @@
+---
+buildpack: https://github.com/cloudfoundry/java-buildpack#v4.5.2
+memory: 1G
+path: /target/demo.jar
+env:
+  JBP_CONFIG_OPEN_JDK_JRE: '{ jre: { version: 1.8.0_+ } }'
+
+#applications:
```
shakuzen: If you uncomment these lines, the test will pass, but these lines shouldn't be needed, I think.

shakuzen:
I have started the process internally to be added to my employer's GitHub organization so I'll be covered by the corporate CLA.

shakuzen:
I have been added to the GitHub organization now. Could you please close and reopen the pull request so the CLA check runs again?

shakuzen:
Thanks for the fix, @nebhale

Issue #850, opened 2018-01-23
Return route service bindings from routes().list(...)


danapsimer:
The Route domain class does not contain a property to report the bound route services.  The CF CLI provides this information in the cf routes output for instance

```
cf routes
Getting routes for org XXXXX / space performance-testing as XXXXX ...

space                 host                               domain                  port   path   type   apps                               service
performance-testing   wire-mock-jazz                     apps-np.xxxxxxxxx.com                        wire-mock-jazz
performance-testing   wire-mock                          apps-np.xxxxxxxxx.com                        wire-mock
performance-testing   asdf-mocks                         apps-np.xxxxxxxxx.com                        asdf-mocks                         asdf-routing-service
performance-testing   asdf-messaging-performance         apps-np.xxxxxxxxx.com                        asdf-messaging-performance
performance-testing   asdf-mocks-a                       apps-np.xxxxxxxxx.com                        asdf-mocks
performance-testing   asdf-routing-service-performance   apps-np.xxxxxxxxx.com                        asdf-routing-service-performance
performance-testing   asdf-order-performance             apps-np.xxxxxxxxx.com                        asdf-order-performance
```
You can see it on the "asdf-mocks" line in the last column.  Where in the API can this information be found?

twoseat:
It looks like we've fallen out of sync with the cli, so I'll look at getting us updated shortly - thanks for the notice!

PR #852, opened 2018-01-30
Add route service name to routes().list() output

danapsimer:
Addresses #850


twoseat:
Thanks for the PR @danapsimer - I'd already got as far as testing our own solution before getting distracted by some backwards compatibility updates, so won't be using your PR (in part because we're currently making all updates to both `2.x` and `master` for now)

The change will be included in the next release (along with those compatibility updates), which I expect today barring problems. And thanks again for reporting this!

PR #861, opened 2018-02-28
Add underscores to variable

kmj940101:
Underscores are missing in the space quota definition guid variable.


twoseat:
Thanks for the PR, @kmj940101 (and forgive us for letting such an obvious mistake stand for so long!). I incorporated the change into our 2.x branch as well.

PR #874, opened 2018-04-04
Update service broker

arghya88:
This adds support for updating service broker using cloudfoundry-operations


twoseat:
Thanks for the PR @arghya88 - unfortunately I forgot to tag it so that this would close automatically, but it's currently working through CI and will be in the next release. I made some [polishing changes](https://github.com/cloudfoundry/cf-java-client/commit/671378e332e7a5c0731fd03cb8a56356356d4fab), that you may want to check out, but most of them were just using IntelliJ's auto-formatting to make things consistent.

arghya88:
@twoseat Can you help me with what setting  I need to put in my IntelliJ so that the formatting looks good for next PR?

PR #882, opened 2018-04-26
ApplicationManifest YAML serialization fixes

### LINKED ISSUE #881: Minor issues in `ApplicationManifestUtils::toYaml`
There should be a null check around the `health-check-type` key, otherwise an NPE is thrown when a type isn't set:
https://github.com/cloudfoundry/cf-java-client/blob/516b00fe3ae989bde68d498853c368965d19cf0b/cloudfoundry-operations/src/main/java/org/cloudfoundry/operations/applications/ApplicationManifestUtils.java#L346

Also, on OSX the YAML serializer doesn't render the path correctly. There should probably be a `toString` call here:
https://github.com/cloudfoundry/cf-java-client/blob/516b00fe3ae989bde68d498853c368965d19cf0b/cloudfoundry-operations/src/main/java/org/cloudfoundry/operations/applications/ApplicationManifestUtils.java#L353
```
---
applications:
- health-check-type: port
  name: whatever
  path: !!sun.nio.fs.UnixPath {}
```

Fitzoh:
fixes #881


Fitzoh:
CLA signed, closing and reopening

PR #890, opened 2018-05-21
Adds space_guid, organization_guid as parameter to the list events

### LINKED ISSUE #889: Additional filters for `/v2/events`
`ListEventsRequest` currently has filters for `actee`, `timestamp`, and `type`.

Add support for `space_guid` and `organization_guid`.

https://apidocs.cloudfoundry.org/1.34.0/events/list_all_events.html

```
Valid filters: timestamp, type, actee, space_guid, organization_guid
```

arghya88:
closes #889

### DIFF
--- a/cloudfoundry-client/src/main/java/org/cloudfoundry/client/v2/events/_ListEventsRequest.java
+++ b/cloudfoundry-client/src/main/java/org/cloudfoundry/client/v2/events/_ListEventsRequest.java
@@ -37,6 +37,20 @@ abstract class _ListEventsRequest extends PaginatedRequest {
     abstract List<String> getActees();
 
     /**
+     * The organization id
+     */
+    @FilterParameter("organization_guid")
+    @Nullable
+    abstract List<String> getOrganizationId();
+
+    /**
+     * The space id
+     */
+    @FilterParameter("space_guid")
+    @Nullable
+    abstract List<String> getSpaceId();
+
+    /**
      * The timestamps
      */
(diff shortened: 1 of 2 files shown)


twoseat:
Thanks @arghya88 - note the pluralisation of parameters, and the error in the integration test (though I understand that's hard to check without a test environment!)

arghya88:
@twoseat I looked at the integration test...what error you got and why you had to do this change if you can explain ?..I have a cf but I am not sure how/if I can run only this test.Tried running the whole set of integration tests but some other test related to buildpack keeps failing..

arghya88:
@twoseat Noted on this...Can you explain what failed and why the reordering the tests were necessary?

I just tried running this test and it fails with below error on my cf instance:

Found unexpected property stack in payload for org.cloudfoundry.client.v2.buildpacks.BuildpackEntity$Json

Any idea?

twoseat:
Hi @arghya88!

The `organisation` test failed because it was trying to filter on `.space()` (https://github.com/cloudfoundry/cf-java-client/pull/890/commits/fe555ec31e2239ef57c75673f8646927d4b1dada#diff-52c358d0d9df26bdda7d2892da39173dR135).

The reordering of tests isn't strictly necessary, but we try to keep everything in alphabetical order (there are exceptions, but not many!). Fortunately hitting alt-cmd-L inside a class in Intellij generally takes care of such things.

The error you're seeing typically means that CF has returned a field that we don't know about. A quick check of the docs suggests that `stack` isn't something we should be expecting, so not sure what's happening...

PR #916, opened 2018-10-25
Support checking client tokens

mheath:
Client tokens do not include these claims: authorizationTime, email,
origin, userId, userName. So these fields need to be made nullable.
a1c8a17


twoseat:
Thanks for the PR @mheath - when validating I noticed that there's another optional that applies to user tokens, so I added that in too.

PR #920, opened 2018-10-31
Add support for user provide service tags

mheath:
API version 2.104.0 added support for user provider service instance
tags. This commit adds support for creating, updating, and querying UPS
tags in Reactor client and operations.

I don't have an environment where I can run the integration tests so I
didn't modify them but I did do some manual testing against PWS
and everything worked.


twoseat:
Thanks for the PR @mheath - I'll need to create an integration test for this (I understand that's the hardest part for contributors to do, so no worries!) so it might take a day or two to get to this, but it will be done before I cut a release.

mheath:
@twoseat I made the fixes you asked for. Let me know if you need anything else.

PR #924, opened 2018-11-20
Added login_hint for Password Grant Token issue 923

pavellom:


pavellom:
Hi @twoseat,

Can you please review this change?

cloudfoundry-client-reactor/src/main/java/org/cloudfoundry/reactor/tokenprovider/_PasswordGrantTokenProvider.java
```diff
@@ -36,6 +36,11 @@
      * The username
      */
     abstract String getUsername();
+    
+    /**
+     * The login_hint
```
twoseat: Javadoc should use English, e.g. "The identity provider to be used", or at least "The login hint"

cloudfoundry-client-reactor/src/main/java/org/cloudfoundry/reactor/tokenprovider/_PasswordGrantTokenProvider.java
```diff
@@ -36,6 +36,11 @@
...
+     */
+    abstract String getLoginHint();
```
twoseat: Field should be marked `Nullable`, as it's optional

twoseat: I'm unsure about this, but I don't think we should treat this as a simple string. If we do that it relies on the user to provide a correctly encoded json object, when they could provide an endpoint (still as a string) that we convert to a correctly encoded json object for them. So the difference between them providing 

`%257B%2522origin%2522%253A%2522uaa%2522%257D`

rather than 

`uaa`

I'd be interested in your thoughts, and those of @andreaAlkalay as users.

twoseat: To highlight the potential confusion that first example might actually need to be `%7B%22origin%22%3A%22uaa%22%7D`, I'm still wrapping my head around that.

pavellom: Hi,

This was exactly my thought at first, but then I thought, what would happen if another parameter will be needed in the wrapped json? What if the existing parameter wiil be renamed from origin to (I don't know) TheOrigin?
I think if UAA asked to provide json as value to parameter called login_hint and not added parameter called origin, is an indication that this is not final, 
and login_hint value could be extended for other purposes other than origin.
If this is the case, I'm not sure that constructing json by ourself rather then asking to provide it as is, is something we would like to handle.

andreaAlkalay: Hi,

I agree with Pavel, the solution he suggests is aligned with the UAA implementation and provides an accurate implementation in the CF client.

Thanks a lot,
Andrea.

torsten-sap: @twoseat Regarding the encoding you are right, it should be %7B%22origin%22%3A%22uaa%22%7D. For some reason the UAA API documentation seems to be wrong (double-encoded) here. @tack-sap and myself (we developed the feature in the UAA) will check and correct it on the UAA API doc side.

The idea was indeed to pass a JSON as login_hint so that other login hints (apart from origin) could still be added later on.

twoseat: Thanks @torsten-sap that's reassuring to know - it seems the example shows an encoded json url inside an encoded url, hence my uncertainty about where the double-encode was coming from. Either way the doc would benefit from a clearer example!

pavellom: Hi @twoseat,

Are you ok with this change?
If so, when do you expect it to be released?

twoseat: The change looks good, I just need to expand it to include the other occurrences of `login_hint`. With luck I should be able to include that tomorrow, so you should see a snapshot build soon.

twoseat:
This was merged in 68ec16bcea71598c7a172238abc5063d260d90e9, I'm not sure why the PR wasn't automatically closed.

PR #935, opened 2019-01-02
Extend create service interface

zlatozar:
Additional parameters could be added
https://apidocs.cloudfoundry.org/6.8.0/service_instances/creating_a_service_instance.htm


twoseat:
Hi @zlatozar The `1.x` has been EOL since `2.x` was introduced, and we just ceased development on `2.x` in favour of the well-established `3.x` line. As a result we're only accepting PRs against 3.x, which I urge you to upgrade to if you possibly can.

Issue #939, opened 2019-01-17
Create Orgs/Spaces Fails When Using Client Credentials


twoseat:
Report from a customer:

"We’re trying to use client credentials auth everywhere as opposed to CF users. I hit a problem when trying to create an Org after making that change. The java client attempts to set the org manager user, by extracting the username of the currently authenticated user from the access token. Only now we’re using a client rather than a user, there is no username available…"

This issue has already been [resolved](resolved) in the CLI, we need to reflect their [solution](solution). Note that there are [other related issues](other) that will probably need handling separately.

[resolved]: https://github.com/cloudfoundry/cli/issues/1418
[solution]: https://www.pivotaltracker.com/n/projects/892938/stories/158986640
[others]: https://www.pivotaltracker.com/n/projects/892938/epics/4056625

Zavael:
HI, I am experiencing again this problem. when using `ClientCredentialsGrantTokenProvider` for `ReactorUaaClient` to creato organization
```java

    @Value("${cf.clientId}")
    private String clientId;

    @Value("${cf.clientSecret}")
    private String clientSecret;

//...

cloudFoundryOperations = DefaultCloudFoundryOperations.builder()
                    .cloudFoundryClient(useCloudFoundryClient())
                    .uaaClient(ReactorUaaClient.builder()
                            .connectionContext(DefaultConnectionContext.builder()
                                    .apiHost(apiHost)
                                    .build())
                            .tokenProvider(ClientCredentialsGrantTokenProvider.builder()
                                    .clientId(clientId)
                                    .clientSecret(clientSecret)
                                    .build())
                            .build())
                    .build();
```

I get this exception:

```
java.lang.IllegalStateException: Unable to retrieve username from token
        at org.cloudfoundry.reactor.uaa.UsernameProvider.lambda$getUsername$3(UsernameProvider.java:75) ~[cloudfoundry-client-reactor-5.1.0.RELEASE.jar:na]
        Suppressed: reactor.core.publisher.FluxOnAssembly$OnAssemblyException:
Assembly trace from producer [reactor.core.publisher.MonoFlatMap] :
        reactor.core.publisher.Mono.checkpoint(Mono.java:1879)
        org.cloudfoundry.operations.organizations.DefaultOrganizations.create(DefaultOrganizations.java:87)
Error has been observed at the following site(s):
[...]

PR #947, opened 2019-02-09
Filter assign users request on creating organization or space, in ca…

pavellom:
https://github.com/cloudfoundry/cf-java-client/issues/939


pavellom:
@twoseat, can you review this PR?

twoseat:
Thanks @pavellom, looking at this now.

twoseat:
Hi @pavellom My concern with this PR is that it's not in line with the solution the CLI team have implemented - we shouldn't just be skipping an association, we should be making it based on the uaa id (see the now-fixed links in the original issue for more detail). So while this fix might work for your specific case, it's not a generalisable solution. I'd be interested to hear your thoughts.

pavellom:
Hi @twoseat,

I agree, I've missed the solution provided by cli colleagues.

PR #948, opened 2019-02-15
Uaa token provider

tareqhs:
Current it is complicated to extend or mock ``AbstractUaaTokenProvider`` because it is locked by ``final`` and package-protected methods. Add an intermediate interface for UAA token providers to allow extending and mocking the UAA functionality more elegantly.

### DIFF
--- a/cloudfoundry-client-reactor/src/main/java/org/cloudfoundry/reactor/UaaTokenProvider.java
+++ b/cloudfoundry-client-reactor/src/main/java/org/cloudfoundry/reactor/UaaTokenProvider.java
@@ -1,0 +1,34 @@
+/*
+ * Copyright 2013-2019 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


nebhale:
Can you give an example of extension and mocking that you're trying to do where the `TokenProvider` interface isn't sufficient?  The class is designed for extension and locks down certain methods that _shouldn't_ be overridden as they define the core behavior of the class.

tareqhs:
@nebhale We have a solution that makes use of UAA-specific token data, e.g. refresh tokens. Writing wrappers and unit test is overly complicated because ``getToken()`` is locked with ``final`` and forces me to go through ``HttpClient`` flows. Also the two abstract methods are package-protected and force me to put the relevant code inside a package that is irrelevant to project ( ``org.cloudfoundry.reactor.tokenprovider`` ).

Locking the code that much makes it actually less extensible.

Kehrlann:
Closing as stale.

PR #982, opened 2019-08-06
Minor improvements for development

nictas:
These improvements would mainly benefit people that develop on Windows or that use Eclipse instead of IntelliJ. Additionally, this pull request adds important information about building the project in an IDE, which is not obvious for people that haven't used Immutables in the past (like myself).


.gitignore
```diff
@@ -2,3 +2,8 @@
 .idea/tasks.xml
 .idea/workspace.xml
 target
+bin
```
twoseat: I'm no expert in .gitignore + Eclipse + Windows, but the internet seems to think that this should be `bin/` (or possibly `/bin/` or `bin/**`). Can you confirm this is correct?

nictas: I just followed the entry you have on line 4. With it you're ignoring all "target" directories by just specifying `target`. That seems to work fine and `bin` works as well - I tested it.

But you're right - `bin/` is the recommended way, so I'll update the pull request. I hope you won't mind modifying the `target` entry to follow suit as well.

nictas:
Included with https://github.com/cloudfoundry/cf-java-client/commit/76c37242d346e37d1b2a8cbe943a059b5f0120ac. Thanks @twoseat!

Issue #938, opened 2019-01-10
Incompatibility with Spring WebFlux 5.1 WebClient


scottfrederick:
When CF Java Client is used in an application that also uses `WebClient` from Spring WebFlux `2.1.x.RELEASE`, exceptions are thrown from Netty and connections to CF fail. 

The relevant exception is: 
```
Caused by: java.lang.IllegalArgumentException: 'httpRedirects' is already in use
	at io.netty.util.ConstantPool.createOrThrow(ConstantPool.java:113) ~[netty-common-4.1.31.Final.jar:4.1.31.Final]
	at io.netty.util.ConstantPool.newInstance(ConstantPool.java:95) ~[netty-common-4.1.31.Final.jar:4.1.31.Final]
	at io.netty.util.AttributeKey.newInstance(AttributeKey.java:55) ~[netty-common-4.1.31.Final.jar:4.1.31.Final]
	at reactor.netty.http.client.HttpClientOperations.<clinit>(HttpClientOperations.java:720) ~[reactor-netty-0.8.3.RELEASE.jar:0.8.3.RELEASE]
```

The root of the issue is clashing versions of `reactor-netty`. 

* CFJC `3.x` depends on `io.projectreactor.ipc:reactor-netty:0.7.x.RELEASE` (via `spring-boot-dependencies:2.0.x`). 
* Spring WebFlux `2.1.x` depends on `io.projectreactor.netty.reactor-netty:0.8.x.RELEASE`. 

These versions of `reactor-netty` both have a class with a static initializer that registers a "constant" with Netty: 

* `reactor-netty:0.7.x.RELEASE`: [reactor.ipc.netty.http.client.HttpClientOperations](https://github.com/reactor/reactor-netty/blob/0.7.x/src/main/java/reactor/ipc/netty/http/client/HttpClientOperations.java#L929-L930)
* `reactor-netty:0.8.x.RELEASE`: [reactor.netty.http.client.HttpClientOperations](https://github.com/reactor/reactor-netty/blob/master/src/main/java/reactor/netty/http/client/HttpClientOperations.java#L738-L739)

Whichever of these initializers runs first wins, and the second one fails. When the `HttpClientOperations` used by CFJC fails to initialize, the exception above appears in stdout but the app will start. Subsequent calls to CFJC result in long delays followed by a timeout.

scottfrederick:
[...]

Issue #908, opened 2018-09-02
SSLEngine and NPE failures under high load


tareqhs:
Hi,

I am experiencing TLS-related inconsistencies if a given ``ConnectionContext`` is used by a high number of CF operations frequently. This is what I do:
1. Create a singleton ``ConnectionContext`` with default configuration
1. Create a singleton ``CloudFoundryClient`` with default configuration
1. list organizations: ``client.organizations.list(...).block()`` 1000 times
1. wait 30 seconds and repeat (3)

After two-three iterations, I start getting exceptions:
* javax.net.ssl.SSLException: SSLEngine closed already
* java.lang.NullPointerException: while trying to invoke the method io.netty.handler.codec.http.HttpMethod.name() of a null object loaded from local variable 'method'

violetagg:
Might be related to https://github.com/reactor/reactor-netty/issues/413

mheath:
I don't think this is related to reactor/reactor-netty#413, I have updated to reactor-netty 0.7.14.RELEASE and I still see this issue.

dmikusa-pivotal:
This is a very old issue and I'm not sure if this still applies to recent versions so I'm going to close this out. 

If you are experiencing this or a similar issue using the latest 4.x or 5.x release of the cf-java-client, then please open a new issue and include the [suggested troubleshooting information](https://github.com/cloudfoundry/cf-java-client#troubleshooting). Thanks!

Issue #903, opened 2018-08-02
Netty Leak & App Crash with OutOfDirectMemoryError


pivotal-david-osullivan:
A sample app which creates a `DefaultCloudFoundryOperations`, `Applications` and `LogRequest` object and queries a given app's logs.

When deployed locally and in a CF container and queried via curl, for example, I see the error: `LEAK: ByteBuf.release() was not called before it's garbage-collected.`, and eventually the container crashes with OutOfDirectMemoryError: `io.netty.util.internal.OutOfDirectMemoryError: failed to allocate 16777216 byte(s) of direct memory (used: 402653191, max: 419430400)`. 

This appears to be the same 2 symptoms discussed in https://github.com/cloudfoundry/cf-java-client/issues/806 which were addressed in v2.17.0. I've enabled advanced leak detection on local instance, the output is attached and the hints seem to refer to 'reactor.left.httpCodec' among others.

Versions in use: reactor-core: 3.1.8, reactor-netty: 0.7.8, client: 3.11.0, Buildpack: 4.13

Please let me know if I can provide any more info.

[Stack.log](https://github.com/cloudfoundry/cf-java-client/files/2253549/Stack.log)

David-Pivotal:
Hi Team, 

Could you share any advice?

Regards,
David

violetagg:
@David-Pivotal Did you run it with `-Dio.netty.leakDetection.level=paranoid` also did you enable `reactor.ipc.netty=DEBUG`?
Also any chance to use Netty 4.1.29.Final?
Thanks,
Violeta

pivotal-david-osullivan:
@violetagg The stack.log above was with .level=advanced, please see attached log with .level=paranoid and the Netty debug set. 

It is based on reactor-netty 0.7.8 which has Netty 4.1.25. Will there be a reactor-netty release soon to include the latest Netty v4.1.29?

[cf-client.log](https://github.com/cloudfoundry/cf-java-client/files/2345716/cf-client.log)

violetagg:
@pivotal-david-osullivan yes there will be a release this week. Also track this https://github.com/reactor/reactor-netty/issues/422 , most probably we leak the memory when timeout appears.

violetagg:
@twoseat @nebhale Do you use Flux/Mono#timeout with Reactor Netty. If yes can you apply in CF Java Client a workaround similar to this one https://github.com/spring-cloud/spring-cloud-gateway/pull/540?
Thanks,
[...]

PR #984, opened 2019-08-12
Adopt Reactor Netty 0.8.10

nictas:
Several notes in no particular order:
* The code in this pull request is **not** formatted with the formatter that is used for the rest of the code base, because I couldn't find it. If anyone can link it to me, I'll be happy to re-format the code.
* There is one HUGE problem that is a result of a limitation in Reactor Netty's 0.8.10 version. Upload of files (application binaries, buildpacks, etc.) can lead to `OutOfDirectMemoryError`s if the size of the file exceeds a certain threshold (around 10MBs). This problem is marked with a `FIXME` comment in the `MultipartHttpClientRequest` and is reported to Reactor Netty in https://github.com/reactor/reactor-netty/issues/811. In my opinion, this pull request should NOT be merged until that bug/limitation is fixed and a newer version of the library can be adopted. With that said, the pull request can be reviewed, as the fix for the issue should not affect any other classes besides `MultipartHttpClientRequest`.
* This pull request includes the commits from https://github.com/cloudfoundry/cf-java-client/pull/982.
* I've tested this pull request with the following integration tests:
  * Create an application, upload binaries, bind it to services, start, stop, delete it, assign routes to it, etc.
  * Create services synchronously and asynchronously (with polling of `last_operation`), create service keys.
  * Retrieve information about existing services, applications, space users (developers/auditors/managers) and other entities.
  * Start tasks and monitor them.
  * Test whether the requests/responses are logged when `cloudfoundry-client.request` and `cloudfoundry-client.response` are set to `DEBUG`.
* The adoption of the new version of Reactor Netty required significant refactoring and changes. I hope you don't mind. If you don't like something, let me know, and I'll see whether it could be improved.
* I couldn't find any good way of splitting this commit into several smaller ones, so apologies for the massive change. Hope you'll find the time to review it. :)
* This pull request may be a fix for https://github.com/cloudfoundry/cf-java-client/issues/903, https://github.com/cloudfoundry/cf-java-client/issues/938 and hopefully https://github.com/cloudfoundry/cf-java-client/issues/908 (which is why I'm doing this pull request in the first place).


twoseat:
I'm testing this change now, and in the meantime have flagged the related reactor-netty issue so hopefully we'll get a fix soon.

nictas:
Thanks! Let me know if I can help in any way.

nictas:
I ran our load tests against this change and unfortunately, I reproduced another issue, which seems to come from the Netty client. I tracked it here:
https://github.com/netty/netty/issues/9459

The "good" news is that for about 20000 requests, I was able to reproduce it only 3 times.

twoseat:
@nictas I've done an initial test run (takes a while!) and all looks good except for downloading a droplet. I'm pretty sure that's just an artifact of the way I'm running the tests, but if you have time and the ability to try downloading a droplet in your own environment I'd be interested in the result.

nictas:
I'll try it out tomorrow.

violetagg:
> I ran our load tests against this change and unfortunately, I reproduced another issue, which seems to come from the Netty client. I tracked it here:
> [netty/netty#9459](https://github.com/netty/netty/issues/9459)
> 

Why do you use an old Netty version (4.1.29). Reactor Netty 0.8.10 runs with 4.1.36 (even pretty well with 4.1.38 except this issue https://github.com/netty/netty/pull/9436, which I believe does not affect you)

nictas:
> Why do you use an old Netty version (4.1.29). Reactor Netty 0.8.10 runs with 4.1.36 (even pretty well with 4.1.38 except this issue netty/netty#9436, which I believe does not affect you)

I hadn't noticed that the Netty client's version is being downgraded until I saw your comment. Apparently, the version of Netty is not coming from Reactor Netty's transitive dependency, but from the `spring-boot-dependencies` POM, which this project inherits. I'll run our load tests again with Netty 4.1.38, but I don't see any mention of the NPE I encountered in their release notes. Fingers crossed though.

twoseat:
For reference I've updated locally to Boot 2.0.9, which provides netty 4.1.34 (I'm also looking at Boot 2.1, but that's a little more involved). Seems to work ok, though I don't know about the impact on the load test issue.

nictas:
Apparently, the Netty issue will be fixed in their next release:
https://github.com/netty/netty/commit/23cf85f6f7ba539be997a38c718cb9ca0f49caaa

nictas:
@twoseat I reproduced the issue with the download of droplets. The problem is that the client does not follow redirects and even when it does, it forwards the `Authorization` header, which leads to `400 Bad Request`s. I'll fix it ASAP.

nictas:
Maybe I'm doing something wrong, but I can't seem to convince the Reactor Netty client to stop forwarding the `Authorization` header. I've opened this issue:
https://github.com/reactor/reactor-netty/issues/818
I'll update this pull request when I have a working solution or a workaround.

nictas:
@twoseat I updated the pull request with my latest changes. The issue with the download of droplets and applications should be fixed now. I tried to download an application, a droplet and a package and all seemed good with one exception - I could not open the downloaded droplet. The file had a significant size of around 300MB, but apparently was not in a ZIP format like the application and the package. Is this normal?

nictas:
Btw, we still should NOT merge this pull request yet. The upload of applications is still a problem for applications larger than 10MBs. @violetagg gave me a workaround for https://github.com/reactor/reactor-netty/issues/811, but I'm having some problems with applying it to the cf-java-client. We're currently looking into what's going wrong and I'll update this pull request as soon as we have a solution.

twoseat:
I believe the droplet is a `.tgz` file, did you try that?

nictas:
You're right! The droplet is in fact a `.tgz` file. Well in that case, everything works at least as far as I can see (application upload excluded, of course). You can re-run your tests if you want.

twoseat:
Thanks for the update (and the ongoing great work!) I'll take a look as soon as I have time. Have you tried using the `0.9.x` RC? Depending on when netty release the fix we're waiting for it's possible we could move straight to `0.9.x`.

nictas:
@twoseat In terms of implementation, this PR is finished (at least until your review - I'll rework it, if you don't like how something turned out). I tested how the file upload works with reactor-netty 0.9.0.RC1 and everything's good at least so far. The only thing left from my side is to run our load tests against this change. I'll post results on Monday (probably).

scottfrederick:
Thanks for the work on this @nictas as @twoseat. I know several teams who are anxious to see this working!

nictas:
@twoseat All of our load and functional tests pass, so from my point-of-view this pull request can now be merged (assuming that you have no comments that need to be addressed). We could also wait for the official 0.9.0 release of reactor-netty.

One important note: We only use a subset of the cf-java-client's features, so our tests can't confirm that **everything** is working properly. I trust your integration tests to catch any issues that have slipped by our tests.

twoseat:
Thanks @nictas - we'll wait for the official `0.9.0` release, but I'll start testing on this version.

violetagg:
You should be able to update this PR to `Reactor Dysprosium GA`

royclarkson:
@nictas and @twoseat I'm adding my thanks and 👍 for the work on this too. Really appreciate it.

Issue #950, opened 2019-02-26
Readme: Bad link for httpClient (Project Reactor)


jfsc:
At :

> `cloudfoundry-client-reactor`  – The default implementati ...

  The link for **HttpClient** `s API is broken.

twoseat:
Fixed in 3e03ed5b2f4d0c5028ede16a2c4f995ce0e6b46c

PR #993, opened 2019-08-16
Fix link in README.md

nictas:
https://github.com/cloudfoundry/cf-java-client/issues/950


twoseat:
I'd already changed this (though to a slightly different url) in an earlier commit. I'll reassess all the links when we merge the netty update, so I'm going to close this.

Issue #913, opened 2018-10-02
Cloudfoundry Operations : Enable service access enabling access for all plans even with plan name specified


arghya88:
Cloudfoundry Operations "enable service access"  API enabling access for all plans even with plan name specified

arghya88:
@twoseat if you have not started already I can start on it.

twoseat:
Haven't started yet, and always happy to accept a PR!

arghya88:
@twoseat..can you prioritize this

twoseat:
@arghya88 Can you clarify what the exact problem is? Specifically do you mean that if you have serviceA and serviceB, and enable service access for serviceA by name, it will enable service access for serviceB?

PR #994, opened 2019-08-16
Fix enabling of public visibility for a single plan

nictas:
Resolves https://github.com/cloudfoundry/cf-java-client/issues/913.


nictas:
@twoseat I've only tested this via unit tests, so far. Will it be a problem for you to test it manually on your side?

twoseat:
Almost certainly I'll create an integration test for it, so not a problem.

nictas:
Thanks! The unit test I've written is a bit clumsy, but I tried to follow the pattern in the test class as much as possible. It would've required significant refactoring to make the test cleaner and still keep it consistent with the rest of the tests in that class.

PR #1029, opened 2020-01-24
Fix indefinite retry on invalid token - configurable retry

valentinEmpy:
Currently, on an InvalidTokenException the Operator retries the request indefinitely, which might be problematic as it'd lead to other issues if abused - IP based (unauthenticated requests) Rate Limit Exceeded for example.

### DIFF
--- a/cloudfoundry-client-reactor/src/main/java/org/cloudfoundry/reactor/ConnectionContext.java
+++ b/cloudfoundry-client-reactor/src/main/java/org/cloudfoundry/reactor/ConnectionContext.java
@@ -56,4 +56,9 @@ public interface ConnectionContext {
      */
     Mono<Void> trust(String host, int port);
 
+    /**
+     * The number of retries after an unsuccessful request
+     */
+    Long getInvalidTokenRetries();
+
 }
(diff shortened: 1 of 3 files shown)


checketts:
I can confirm that this fix seems to have addressed #1032 

I've run for 3 days with it and before I would see at least 2 daily issues. Now there are no memory issues. 

(The request that triggers the refresh now can fail and retry on its own instead of the application using up all its memory) Looking forward to a release (both on the v3 and v4 lines)

checketts:
@twoseat Would it be possible to release this soon? This has been plaguing [Promregator](https://github.com/promregator/promregator) users for months and we have now verified with several users this fix is working for them.  I would love to update to a released version of the library so we can roll out the fix to more folks.

twoseat:
I was holding a release pending the thumbs up on another fix. I'll chase that now, and unless I hear something bad from that I'll do a sweep of dependency updates and release this week.

eaglerainbow:
@twoseat Thanks! I am on standby for your next release to incorporate it and get this ugly thing out of the way for [us](https://github.com/promregator/promregator).

twoseat:
The other fix is OK, slight delay on my end for various reasons but I am working on the release, so should be out today or tomorrow! Thanks for your patience.

checketts:
Thanks again. I see the release tag was created. (https://github.com/cloudfoundry/cf-java-client/tree/v4.4.0.RELEASE)

But it isn't in Maven central yet. https://mvnrepository.com/artifact/org.cloudfoundry/cloudfoundry-client-reactor (Sorry if I'm jumping the gun) Does it usually take a while to propogate?

twoseat:
Some issues with oout CI publishing process that I'm currently working on - the `3.x` line has made it out, `4.x` should follow in a couple of hours, fingers crossed!

PR #1033, opened 2020-02-12
Fix Metadata serialization (serialize null values)

valentinEmpy:
As per documentation in order to delete labels/annotations from a resource's metadata, one must:

> To delete, include the existing key with a null value.

Explained in: https://v3-apidocs.cloudfoundry.org/version/3.80.0/index.html#Using-labels-and-annotations

The Immutables class includes `AllowNulls` annotation but **does not have** `JsonInclude`. This makes the default `ObjectMapper` which is initialized using `setSerializationInclusion(NON_NULL)` to IGNORE the null mappings in the label/annotation maps and does not serialize them when sending the request to the controller. This in turn, does not allow the deletion of Metadata keys at all.
Using `JsonInclude` annotation overriding the default serialization inclusion handles this issue.

### DIFF
--- a/cloudfoundry-client/src/main/java/org/cloudfoundry/client/v3/_Metadata.java
+++ b/cloudfoundry-client/src/main/java/org/cloudfoundry/client/v3/_Metadata.java
@@ -16,6 +16,7 @@
 
 package org.cloudfoundry.client.v3;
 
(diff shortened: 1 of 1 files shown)


valentinEmpy:
There's the possibility to provide a custom `ObjectMapper` with a different default serialization inclusion but this would lead to side effects of other calls we do.

Issue #1034, opened 2020-02-17
Plus signs are not encoded


nictas:
Plus signs are interpreted as whitespaces by the CF controller. The CLI commands executed below clearly demonstrate this (note the GUID of the space returned by `cf curl`:
```
$ cf space "foo+bar" --guid
2c094d32-4c01-460b-94d0-f73484850d1f
$ cf space "foo bar" --guid
34f23458-0456-4dcf-b15c-a60474bef3ec
$ cf curl "/v2/spaces?q=name:foo+bar"
{
   "total_results": 1,
   "total_pages": 1,
   "prev_url": null,
   "next_url": null,
   "resources": [
      {
         "metadata": {
            "guid": "34f23458-0456-4dcf-b15c-a60474bef3ec",
            "url": "/v2/spaces/34f23458-0456-4dcf-b15c-a60474bef3ec",
            "created_at": "2020-01-30T11:09:34Z",
            "updated_at": "2020-01-30T11:09:34Z"
         },
         "entity": {
            "name": "foo bar",
            ...
         }
      }
   ]
}
```
Due to this, "+" should be encoded to "%2C" by the CF Java client, in order to keep the intention of the request unchanged. This is not done, however, as can be seen in the output of this test:
```
import org.cloudfoundry.client.CloudFoundryClient;
import org.cloudfoundry.client.v2.spaces.CreateSpaceRequest;
import org.cloudfoundry.client.v2.spaces.CreateSpaceResponse;
import org.cloudfoundry.client.v2.spaces.ListSpacesRequest;
import org.cloudfoundry.client.v2.spaces.ListSpacesResponse;
import org.cloudfoundry.client.v2.spaces.SpaceResource;
import org.cloudfoundry.reactor.ConnectionContext;
import org.cloudfoundry.reactor.DefaultConnectionContext;
import org.cloudfoundry.reactor.TokenProvider;
import org.cloudfoundry.reactor.client.ReactorCloudFoundryClient;

import reactor.core.publisher.Mono;

public class PlusEncoding {

    private static final String OAUTH_TOKEN = "bearer <TOKEN>"; // Set your OAuth token here (retrievable via "cf oauth-token").
    private static final String ORGANIZATION_GUID = "<ORGANIZATION_GUID>"; // Set the GUID of an organization in which you can create spaces.
    private static final String API_HOST = "<API_HOST>"; // Set an API host here.
    
    public static void main(String[] args) {
        CloudFoundryClient client = createClient();
[...]

PR #1036, opened 2020-02-19
Fix encoding of plus sign in query params

nictas:
Fixes https://github.com/cloudfoundry/cf-java-client/issues/1034.


cloudfoundry-client-reactor/src/test/java/org/cloudfoundry/reactor/uaa/authorizations/ReactorAuthorizationsTest.java
```diff
@@ -97,7 +97,7 @@ public void authorizeByAuthorizationCodeGrantBrowser() {
     public void authorizeByAuthorizationCodeGrantHybrid() {
         mockRequest(InteractionContext.builder()
             .request(TestRequest.builder()
-                .method(GET).path("/oauth/authorize?client_id=app&redirect_uri=http://localhost:8080/app/&scope=openid&response_type=code%20id_token")
+                .method(GET).path("/oauth/authorize?client_id=app&redirect_uri=http%3A%2F%2Flocalhost%3A8080%2Fapp%2F&scope=openid&response_type=code%20id_token")
```
nictas: I don't particularly like how this turned out, but tests like this one already had some encoded characters even before my change. See `org.cloudfoundry.reactor.uaa.clients.ReactorClientsTest.list()`.

cloudfoundry-client-reactor/src/main/java/org/cloudfoundry/reactor/util/UriQueryParameters.java
```diff
@@ -0,0 +1,36 @@
...
+
+import org.springframework.web.util.UriComponentsBuilder;
+
+import java.util.List;
+import java.util.stream.Stream;
+
+public class UriQueryParameters {
+
+    public static void set(UriComponentsBuilder builder, Stream<UriQueryParameter> uriQueryParameters) {
+        // Replace all literal values with URI variables to apply more strict encoding:
```
twoseat, line 27: We rarely include comments as it's often a sign that the code itself is obscure, or a method within it could be named better. Having said that I don't see anything here that should obviously be changed, but if you have any ideas feel free!

cloudfoundry-client-reactor/src/main/java/org/cloudfoundry/reactor/util/UriVariablesRegistry.java
```diff
@@ -0,0 +1,42 @@
...
+    private final List<UriVariable> uriVariables = new ArrayList<>();
+
+    public Map<String, Object> getUriVariablesMap() {
+        return uriVariables.stream().collect(Collectors.toMap(UriVariable::getKey, UriVariable::getValue));
+    }
+
+    public UriVariable register(Object value) {
+        UriVariable uriVariable = UriVariable.of(getNextUriVariableKey(), value);
+        uriVariables.add(uriVariable);
+        return uriVariable;
```
twoseat, line 35: House style is a blank line above return statements if they're not one-liners.

cloudfoundry-client-reactor/src/main/java/org/cloudfoundry/reactor/util/_UriVariable.java
```diff
@@ -0,0 +1,37 @@
...
+ */
+
+package org.cloudfoundry.reactor.util;
+
+import org.immutables.value.Value;
+
+@Value.Immutable
+public interface _UriVariable {
+
+    String PLACEHOLDER_PATTERN = "{%s}";
```
twoseat, line 24: Any reason this shouldn't be psf?

nictas: It is by default, since this is an interface.

cloudfoundry-client-reactor/src/main/java/org/cloudfoundry/reactor/util/UriQueryParameters.java
nictas: I also prefer to not include comments, but I think this one will be helpful, since it's not immediately obvious why we're replacing literal values with URI variables.

PR #1059, opened 2020-08-11
fixed comparison when http response is Unauthorized

mmueller90:
Hello,

I've noticed using the cf-java-client, using the ClientCredentialsTokenProvider that the access token gets never invalidated after the token is expired. After some investigation I've found out, that 401 Responses from the UAA (for my usage) are not correctly handled. That leads to that the token is never invalidated and already expired tokens are used to access the API.
These PR simply fixes the comparison for isUnauthorized using equals instead of "==".

Best Regards
Marcel


jamesdobson:
Hi @mmueller90 , this is affecting me as well; thanks for making a PR.

Are you able to sign the CLA? If not, may I submit my own PR with the fix?

mmueller90:
> Hi @mmueller90 , this is affecting me as well; thanks for making a PR.
> 
> Are you able to sign the CLA? If not, may I submit my own PR with the fix?

hi @jamesdobson, unfortunately I'm not able to sign the PR so you can submit your own PR with the fix

Issue #1067, opened 2020-09-08
Add support for java9+


gberche-orange:
## expected behavior

As a cf-java-client user 
* in order to leverage java9+ language features
* I need cf-java-client to be compatible with [java9+ LTS version](https://en.wikipedia.org/wiki/Java_version_history), such as java11

## observed behavior

* cf-java-client seems only tested in its ci against java8
https://github.com/cloudfoundry/cf-java-client/blob/a83ea56d7adacac42f99a3b60ad98b84e03c14c1/ci/unit-test.yml#L4-L8
* cf-java-client running with java14 fails with the following stack trace, with root cause https://github.com/jwtk/jjwt/issues/333#issuecomment-409754833 fixed in 0.10.0 while cf-java-client still uses 0.9.1

https://github.com/cloudfoundry/cf-java-client/blob/c4b037e81f2fa06bf650ae8aa11d4ad6544f2429/pom.xml#L66

```
    2020-09-07T19:39:30.07+0200 [APP/PROC/WEB/1] OUT 2020-09-07 17:39:30.078  WARN 19 --- [-client-epoll-1] i.n.c.AbstractChannelHandlerContext      : An exception 'java.lang.NoClassDefFoundError: javax/xml/bind/DatatypeConverter' [enable DEBUG level for full stacktrace] was thrown by a user handler's exceptionCaught() method while handling the following exception:
     2020-09-07T19:39:30.07+0200 [APP/PROC/WEB/1] OUT java.lang.NoClassDefFoundError: javax/xml/bind/DatatypeConverter
     2020-09-07T19:39:30.07+0200 [APP/PROC/WEB/1] OUT 	at io.jsonwebtoken.impl.Base64Codec.decode(Base64Codec.java:26) ~[jjwt-0.9.1.jar:0.9.1]
     2020-09-07T19:39:30.07+0200 [APP/PROC/WEB/1] OUT 	at io.jsonwebtoken.impl.Base64UrlCodec.decode(Base64UrlCodec.java:78) ~[jjwt-0.9.1.jar:0.9.1]
     2020-09-07T19:39:30.07+0200 [APP/PROC/WEB/1] OUT 	at io.jsonwebtoken.impl.AbstractTextCodec.decodeToString(AbstractTextCodec.java:36) ~[jjwt-0.9.1.jar:0.9.1]
[...]

PR #1069, opened 2020-09-23
Update unit test badges for java8 and java11

gberche-orange:
Relates to https://github.com/cloudfoundry/cf-java-client/issues/1067


gberche-orange:
thanks !

PR #1085, opened 2021-01-18
Use long instead of integer

IvanBorislavovDimitrov:
Hello colleagues,

I found a problem with the cf-java-client.
If you create an application with more than 2GB of memory/disk_quota and try to get the process statistics for this application the JSON parsing fails because the controller returns "memory/disk_quota" in bytes and exceeds the integer limits.
Check:
Numeric value (2147483648) out of range of int (-2147483648 - 2147483647)
 at [Source: (byte[])"{
  "resources": [
    {
      "type": "web",
      "index": 0,
      "state": "RUNNING",
      "usage": {
        "time": "2021-01-15T13:46:43+00:00",
        "cpu": 0.0,
        "mem": 285628069,
        "disk": 250437632
      },
      "host": "10.0.137.7",
      "uptime": 2,
      "mem_quota": 2147483648,
      "disk_quota": 1073741824,
      "fds_quota": 16384,
      "isolation_segment": null,
      "details": null,
      "instance_ports": [
        {
          "external": 61028,
          "[truncated 1073 bytes]; line: 15, column: 30] (through reference chain: org.cloudfoundry.client.v3.applications.GetApplicationProcessStatisticsResponse$Json["resources"]->java.util.ArrayList[0]->org.cloudfoundry.client.v3.processes.ProcessStatisticsResource$Json["mem_quota"]) 

This fix is pretty simple and I would be happy to contribute it.


dmikusa-pivotal:
Thanks for the contribution and sorry for the delay. This looks good. I just need to run through the test suite. If that is green, I will merge this. Thanks

dmikusa-pivotal:
Tests pass. Merging.

This will change the ProcessStatistics API, since the function signature is changing from Integer to Long. That needs to be documented in the release notes.

Resolves: #1086

PR #1124, opened 2021-10-25
Allow empty query params for V3 filtering & Fix build for Windows

radito3:
When creating a V3 request with an empty string as a list element, it is ignored and the filter is not applied to the CC API request.
_e.g.:_
```java
CloudFoundryClient client = /* ... */;
client.routesV3()
      .list(ListRoutesRequest.builder()
           .spaceGuid(/* ... */)
           .path("") //this is ignored
           .build())
      .block(); //returns all routes in the space, not filtered by path
```

The CC API allows filtering with empty strings, so it makes sense that the Java client should too.
_e.g.:_ 
The request `<cf-api>/v3/routes?space_guids=<guid>&paths=` is valid

This PR adds support for filtering with empty strings.

It also fixes building on Windows, as some tests were not platform-agnostic and/or broken for Windows.


dmikusa-pivotal:
Sorry, I got pulled away to a different issue.

>As for the second point, when I was building with maven (with jdk 8), it failed on ApplicationManifestUtilsTest because, in one test, the manifest was incorrectly formed and in others, the mock path for an application was being compared in a platform-specific way (it was comparing /some-path with (resolved on windows) C:\some-path).
>Fixing these small test issues enables the build to complete on Windows.

Looks good to me.

>Upon further inspection, I found out that all query filter params that allow for multiple values (all that are plural, e.g. names, paths, space_guids) allow empty values. Or, in terms of code, all that return a Collection.
And all that allow a single value (or singular, e.g. label_selector) don't.
>So, with that in mind, would you reconsider my current solution or would you prefer I add the allowsEmpty method?

OK, perfect. I think that fits with what you submitted in the PR. I'll go ahead and merge this.

PR #1126, opened 2021-11-18
Avoid NPE in DefaultApplications.java

marc-schaefers:
avoid NPE, if environmentJsons is null from application entity

### DIFF
--- a/cloudfoundry-operations/src/main/java/org/cloudfoundry/operations/applications/DefaultApplications.java
+++ b/cloudfoundry-operations/src/main/java/org/cloudfoundry/operations/applications/DefaultApplications.java
@@ -791,7 +791,9 @@ public final class DefaultApplications implements Applications {
             .singleOrEmpty()
             .flatMap(application -> {
                 Map<String, Object> entityEnvironmentJsons = ResourceUtils.getEntity(application).getEnvironmentJsons();
-                Map<String, Object> environmentJsons = new HashMap<>(entityEnvironmentJsons != null ? entityEnvironmentJsons : Collections.emptyMap());
+                Map<String, Object> environmentJsons = Optional
+                    .ofNullable(entityEnvironmentJsons)
+                    .orElse(Collections.emptyMap());
                 Optional.ofNullable(manifest.getEnvironmentVariables()).ifPresent(environmentJsons::putAll);
 
                 return requestUpdateApplication(cloudFoundryClient, ResourceUtils.getId(application), environmentJsons, manifest, stackId)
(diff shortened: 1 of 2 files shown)


dmikusa-pivotal:
@marc-schaefers - Thanks for the submission! Are you seeing this NPE in actual usage of the library? Do you have a stack trace you can share? Just looking for a little more context on this.

Also, you need to click through the links above and appease the EasyCLA bot before we can accept this PR.

marc-schaefers:
Hi, i use the ﻿﻿PushApplicationRequest to push a static build pack with the cloud foundry java libs.
I want to push it on a customer cloud foundry environment, so i haven't configure it or anything.

On one app name i get a nullpointer exception:
`ERROR 18.11.2021 09:22:23.296 (de.aboutcontent.cloudfoundry.deploy.DeployExecutable): error on push
java.lang.NullPointerException
	at org.cloudfoundry.operations.applications.DefaultApplications.lambda$getApplicationId$67(DefaultApplications.java:793)
	Suppressed: reactor.core.publisher.FluxOnAssembly$OnAssemblyException: 
Assembly trace from producer [reactor.core.publisher.MonoIgnoreElements] :
	reactor.core.publisher.Mono.checkpoint(Mono.java:1902)
	org.cloudfoundry.operations.applications.DefaultApplications.pushManifest(DefaultApplications.java:432)
Error has been observed at the following site(s):
	|_ Mono.checkpoint ⇢ at org.cloudfoundry.operations.applications.DefaultApplications.pushManifest(DefaultApplications.java:432)
	|_ Mono.checkpoint ⇢ at org.cloudfoundry.operations.applications.DefaultApplications.push(DefaultApplications.java:404)
Stack trace:
		at org.cloudfoundry.operations.applications.DefaultApplications.lambda$getApplicationId$67(DefaultApplications.java:793)ations.java:793)`


So i have changed this line to insert an empty map, if the jsons env null and now it works like expected.

dmikusa-pivotal:
Thanks for the additional info. 

Two points of feedback on this:

1. Can you add a unit test? [Probably here](https://github.com/cloudfoundry/cf-java-client/blob/main/cloudfoundry-operations/src/test/java/org/cloudfoundry/operations/applications/DefaultApplicationsTest.java). Just to reproduce the error and then verify it's fixed.
2. Could you use `Optional` instead of the ternary operation & `!= null` check? Not a major difference, but it matches the style of the rest of the code better.

Thanks

marc-schaefers:
Hi @dmikusa-pivotal , the source code is updated now and i added a new push test case.

dmikusa-pivotal (approved):

dmikusa-pivotal:
Sorry for the delay, Log4j2 issues had us scrambling before the holidays. Thanks for the PR!

PR #1153, opened 2022-05-25
Add variable substitution in ApplicationManifestUtils

luschmar:
Add functionality [Variables to a Manifest](https://docs.cloudfoundry.org/devguide/deploy-apps/manifest-attributes.html#variable-substitution) in ApplicationManifestUtils

### DIFF
--- a/cloudfoundry-operations/src/main/java/org/cloudfoundry/operations/applications/ApplicationManifestUtils.java
+++ b/cloudfoundry-operations/src/main/java/org/cloudfoundry/operations/applications/ApplicationManifestUtils.java
@@ -42,6 +42,9 @@ import java.util.function.Function;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
+import static java.util.Collections.emptyMap;
+import static java.util.stream.Collectors.toMap;
+
 /**
  * Utilities for dealing with {@link ApplicationManifest}s.  Includes the functionality to transform to and from standard CLI YAML files.
  */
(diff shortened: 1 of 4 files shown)


cloudfoundry-operations/src/test/java/org/cloudfoundry/operations/applications/ApplicationManifestUtilsTest.java
```diff
@@ -455,6 +455,22 @@ public void readSingleBuildpack() throws IOException {
         assertThat(actual).isEqualTo(expected);
     }
 
+    @Test
```
dmikusa-pivotal, line 458: I'd like to see some more test coverage here. 

Off the top of my head:

1. What happens if the manifest.yml has a placeholder that's not given a value? I feel like this should surface some sort of error.
2. What if there are values provided but no placeholder for them?  This seems ignorable.
3. Are there things one could do to accidentally or intentionally break this functionality? I do see that you're quoting the key as it's put into a regex, it would probably be good to have the test do something that would break w/out quoting.

dmikusa-pivotal (changes requested):

cloudfoundry-operations/src/test/java/org/cloudfoundry/operations/applications/ApplicationManifestUtilsTest.java
luschmar: The whole behavior is not specified. My knowledge relies only on this passage [Add Variables to a Manifest](https://docs.cloudfoundry.org/devguide/deploy-apps/manifest-attributes.html#variable-substitution)

1. I do not think; not substituted variables should throw an error. It's no illegal to have an app name like ((abc))
2. I ignore this; 
3. As you can see in some additional tests, there are a lot of edge cases. Since there is no grammar or more specification I cannot preprocess the variable keys or even values.

I know a lot of things can go wrong with this code, when someone tries to rape this functionality. But lets be honest, average users of this substitution will create var-files for every environment with simple text/integer values and use them instead creating multiple manifest for every environment.

My intention for this MR: I have some projects they rely on these substitution stuff used with cf-cli. I want to replace my long bash-script-bloat (blue/green deployment; rolling restart; org cleanup) uses cf-cli and I want some stable cf-java-client replacement. 

Before I rewrite my MR; perhaps you can provide me more insights of this functionality.

Some resources I consulted:
 * https://cli.cloudfoundry.org/en-US/v8/
 * https://v3-apidocs.cloudfoundry.org/version/3.117.0/index.html

luschmar: I changed the approach. I just search for variables and replace them accordingly. And added more tests

cloudfoundry-operations/src/main/java/org/cloudfoundry/operations/applications/ApplicationManifestUtils.java
```diff
@@ -125,21 +149,32 @@ public static void write(OutputStream out, List<ApplicationManifest> application
         }
     }
 
-    private static <T> void as(Map<String, Object> payload, String key, Function<Object, T> mapper, Consumer<T> consumer) {
+    private static <T> void as(Map<String, Object> payload, String key, Map<String, String> variables, Function<Object, T> mapper, Consumer<T> consumer) {
         Optional.ofNullable(payload.get(key))
+            .map(o -> {
+                if(o instanceof String) {
+                    Matcher m = FIND_VARIABLE_REGEX.matcher(((String) o));
```
dmikusa-pivotal, line 156:
```suggestion
                    Matcher m = FIND_VARIABLE_REGEX.matcher((String) o);
```

I don't believe you need the extra `(..)`.

dmikusa-pivotal (approved):

dmikusa-pivotal:
Thanks @luschmar !

PR #1159, opened 2022-07-05
Allow null values in map of service parameters

theghost5800:
We hit scenario where service brokers return response including null values and this leads to parsing exception. This is a client limitation and needs to be fixed.

### DIFF
--- a/cloudfoundry-client/src/main/java/org/cloudfoundry/client/v3/serviceinstances/_GetManagedServiceParametersResponse.java
+++ b/cloudfoundry-client/src/main/java/org/cloudfoundry/client/v3/serviceinstances/_GetManagedServiceParametersResponse.java
@@ -2,6 +2,7 @@ package org.cloudfoundry.client.v3.serviceinstances;
 
 import java.util.Map;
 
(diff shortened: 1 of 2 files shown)


dmikusa-pivotal:
OK, I'm not opposed to the change. Do you have any more details you can share about the circumstances that contribute to this problem? 

It looks like there are cases where the response from server contains a null parameters/credentials data, not an empty map, but it's missing or specifically null? Is that right? I'm just wondering if this is unique to a particular broker? or version of CF? 

Thanks

theghost5800:
Hi @dmikusa-pivotal , We had identified recently that some service brokers in CF returns as response map with keys which can be null and this leads to errors in client and we can't continue further. Previously we were using [v2 alternative](https://github.com/cloudfoundry/cf-java-client/blob/main/cloudfoundry-client/src/main/java/org/cloudfoundry/client/v2/serviceinstances/_GetServiceInstanceParametersResponse.java) which allows null values in map and we don't have any issues. 
So this is missed annotation to be added from my contribution before.

dmikusa-pivotal:
Understood. Thanks for the PR. Merging.

PR #1171, opened 2022-11-11
Fix polling of async v3 jobs and make getLastOperation method public

theghost5800:
The fix related with `waitForCompletion` method is related to state [POLLING ](https://v3-apidocs.cloudfoundry.org/version/3.128.0/#jobs).

Second change is related with ability to use "lastOperation" field from [ServiceBinding](https://github.com/cloudfoundry/cf-java-client/blob/main/cloudfoundry-client/src/main/java/org/cloudfoundry/client/v3/servicebindings/ServiceBinding.java)

### DIFF
--- a/cloudfoundry-client/src/main/java/org/cloudfoundry/client/v3/servicebindings/ServiceBinding.java
+++ b/cloudfoundry-client/src/main/java/org/cloudfoundry/client/v3/servicebindings/ServiceBinding.java
@@ -45,7 +45,7 @@ public abstract class ServiceBinding extends Resource {
      */
     @JsonProperty("last_operation")
     @Nullable
-    abstract LastOperation getLastOperation();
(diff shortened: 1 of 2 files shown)


boyan-velinov (approved):

dmikusa (approved):

IvanBorislavovDimitrov:
Hello @dmikusa,
Are there any plans to merge and release this change?
Best regards,
Ivan

dmikusa:
The CI for this runs on VMware's infrastructure. I no longer work at VMware, so I don't have access anymore to run the integration tests & check if there are issues. @pivotal-david-osullivan needs to do that for the time being.

Last I heard, the CFF is working on setting up some environments so it might be possible over time to move this infrastructure out of VMware to the CFF, where it would be public.

PR #1190, opened 2023-08-03
Fix health check type serialization in manifestV3/process

schulzh:


pivotal-david-osullivan:
Incorporated via #1219

PR #1192, opened 2023-08-28
Mask off non-permission parts of unix file mode for resource match

schulzh:
When requesting the resource match API during pushing of an application, the cloud controller now expects only the permissions part of the file mode to be sent. This change masks off the other part (file type) to restore compatibility.
See https://github.com/cloudfoundry/cloud_controller_ng/pull/3329 for more details.


radoslav-tomov:
Change looks good to me. I would expect the permission to be encoded in short POSIX format anyways.

radoslav-tomov:
@mheath  @dmikusa Can you guys check this out ?

beyhan:
@anthonydahanne thanks for looking into this in #1216. To my understanding this can be closed now because it is applied via #1216 or?

pivotal-david-osullivan:
Closing as work integrated via #1216

PR #1201, opened 2023-10-13
Upgrades to cf-java-client

pacphi:
It's been 7 months since the 5.10.0.RELEASE. The current release documentation states that that release is compatible with Spring Boot release versions up to 2.6.x. Meanwhile, Spring Boot 3.1.x is the current release, with Spring Boot 3.2.x nearing GA.

There's not much that needed to change to prepare a version of the cf-java-client to be compatible with Spring Boot 3.1.x, but it does require setting target to Java 17. That required a change to the maven-compiler-plugin compileArgs configuration to facilitate source code generation and data-binding when using Immutables. I also took the test dependencies forward, so that in the future, the tests could be refactored to leverage JUnit 5 APIs. The rest of the changes are pretty straight-forward, just upgrades to dependencies and plugins to the latest available as of the date of this pull request.

Default build with Maven 3.9.4 targets Java 8 and upgrades Spring Boot to 2.7.16.

Maven profiles get activated based on the installed version of the JDK detected.  These profiles update the dependencies and plugins to support compilation and testing on Java 17 and 21 with Spring Boot 3.1.4 dependency management.

A Github Action workflow was added to provide signal on "good builds".

Currently, tests fail when targeting either Java 17 or 21.  See this open issue for more details: https://github.com/immutables/immutables/issues/1339.

Also did a fair amount of cleanup across Maven child modules where we had unused imports.

> This PR was based upon earlier work in https://github.com/cloudfoundry/cf-java-client/pull/1198.


anthonydahanne:
@pivotal-david-osullivan I think this PR puts us in a good spot: removed unused imports, deps upgrades, and I checked it builds on Java 8

We can address JDK 17 and 21 Immutable issues in a later PR

LGTM, please review! (I don't have approver rights on this project)

pivotal-david-osullivan (approved):

PR #1212, opened 2024-02-07
Fix integration tests

anthonydahanne:


anthonydahanne:
was fixed in this PR https://github.com/cloudfoundry/cf-java-client/pull/1215

PR #1244, opened 2024-10-09
add cnb in LifecycleType

s-yonkov-yonkov:


beyhan:
The CNB lifecycle functionality was added to CF based on https://github.com/cloudfoundry/community/blob/main/toc/rfc/rfc-0028-cnb-lifecycle.md. Details for the implementation are available in https://github.com/cloudfoundry/community/issues/891

PR #1246, opened 2024-10-14
Use yaml instance instead of static variable to enable threadsaftey

### LINKED ISSUE #1245: `ApplicationManifestUtilsV3#write` is not threadsafe
The method `ApplicationManifestUtilsV3#write` uses a static snakeyaml `Yaml` dumper to write all manifests into a byte array this operation is inheritly not threadsafe, as the deserializer is not.

This in itself may not a be problem. However, when using `Applications#pushManifestv3` to push multiple Apps in parallel this can cause issues that manifest in the form of 

- `Bad request: Manifest does not support Anchors and Aliases` or
- `java.lang.NullPointerException: Nodes must be provided.`

LitschiW:
Fixes #1245

### DIFF
--- a/cloudfoundry-operations/src/main/java/org/cloudfoundry/operations/applications/ApplicationManifestUtilsCommon.java
+++ b/cloudfoundry-operations/src/main/java/org/cloudfoundry/operations/applications/ApplicationManifestUtilsCommon.java
@@ -48,14 +48,12 @@ abstract class ApplicationManifestUtilsCommon {
 
     static final int GIBI = 1_024;
 
-    static final Yaml YAML;
+    static final DumperOptions dumperOptions;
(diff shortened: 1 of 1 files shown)


Lokowandtg (approved):
LGTM

Lokowandtg:
We have the same coding in class ApplicationManifestUtilsCommon. Could you apply the changes there as well?

Lokowandtg (approved):
LGTM

anthonydahanne (approved):
looks good to me!

Kehrlann:
@LitschiW Additionally, this does not compile: `ApplicationManifestUtils` uses `ApplicationManifestUtilsCommon#YAML`.

PR #1261, opened 2025-02-11
fix docker serialization and disk_quota

schulzh:
This fixes two issues we've encountered with the manifest serialization.

The first is that docker was not serialized as a deep structure, but as a string (java object representation), which breaks the yaml. The second issue is that the disk quota was not serialized with the correct field name.


anthonydahanne:
Hello @schulzh , thank you for your contribution!
Could you please add a unit test just to assert this does not change inadvertently?

schulzh:
Hi @anthonydahanne , I've added a unit test, could you take a look again?

schulzh:
Hi @anthonydahanne , I've added two more serialization fixes to the PR.

Lokowandtg (commented):
lgtm
(but not tested)

Issue #1242, opened 2024-10-07
cloudfoundry-client: V3.securitygroups _ListSecurityGroupsRequest extends the wrong pagination request


SaifuddinMerchant:
Version 3 List Security Groups is extending the wrong version of pagination request
import org.cloudfoundry.client.v2.PaginatedRequest;

[See _ListSecurityGroupsRequest.java#L17](https://github.com/cloudfoundry/cf-java-client/blob/main/cloudfoundry-client/src/main/java/org/cloudfoundry/client/v3/securitygroups/_ListSecurityGroupsRequest.java#L17)

Error when using the client to make a ListSecurityGroupRequest

```
org.cloudfoundry.client.v3.ClientV3Exception: CF-UnprocessableEntity(10008): Unknown query parameter(s): 'results-per-page'. Valid parameters are: 'page', 'per_page', 'order_by', 'created_ats', 'updated_ats', 'guids', 'names', 'running_space_guids', 'staging_space_guids', 'globally_enabled_running', 'globally_enabled_staging'
	at org.cloudfoundry.reactor.util.ErrorPayloadMappers.lambda$null$2(ErrorPayloadMappers.java:62)
```

The generated class `ListSecurityGroupsRequest` also has the wrong imports, but I am guessing if we fix the import in `_ListSecurityGroupsRequest.java` that will automatically fix the generated code

CF Java Client Version: 5.12.2.RELEASE
CF API version:    3.166.0

Lokowandtg:
Will look into this.

Lokowandtg:
Hi @SaifuddinMerchant, 
can you provide some code to reproduce the error? It would best fit into 
[integration-test/src/test/java/org/cloudfoundry/client/v3/SecurityGroupsTest.java](https://github.com/cloudfoundry/cf-java-client/blob/main/integration-test/src/test/java/org/cloudfoundry/client/v3/SecurityGroupsTest.java)

At least some (failing) skeleton would be great, as I was not able to reproduce your error. 

Thanks, Georg

SaifuddinMerchant:
Hi,

I will provide another comment that will have the test case in the Integration Test format. Here a general Rest API way to reproduce the error

package net.jpmchase.gaia.usage;

import net.jpmchase.gaia.usage.client.CloudFoundryClientFactory;
import org.cloudfoundry.client.CloudFoundryClient;
import org.cloudfoundry.client.v3.securitygroups.ListSecurityGroupsRequest;
[...]

PR #1277, opened 2025-05-09
Fixing pagination on v3._ListSecurityGroupsRequest

SaifuddinMerchant:
v3._ListSecurityGroupsRequest extends the v2.PaginatedRequest which is incorrect. It needs to extend the v3.PaginatedRequest. The code fixes the incorrect package referenced in v3._ListSecurityGroupsRequest and adds a integration test case

Issue Link: https://github.com/cloudfoundry/cf-java-client/issues/1242

### DIFF
--- a/integration-test/src/test/java/org/cloudfoundry/client/v3/SecurityGroupsTest.java
+++ b/integration-test/src/test/java/org/cloudfoundry/client/v3/SecurityGroupsTest.java
@@ -18,6 +18,8 @@ import static org.cloudfoundry.client.v3.securitygroups.Protocol.TCP;
 
 import java.time.Duration;
 import java.util.Arrays;
+import java.util.Collections;
+
 import org.cloudfoundry.AbstractIntegrationTest;
 import org.cloudfoundry.client.CloudFoundryClient;
 import org.cloudfoundry.client.v3.securitygroups.BindRunningSecurityGroupRequest;
@@ -34,6 +36,7 @@ import org.cloudfoundry.client.v3.securitygroups.Rule;
 import org.cloudfoundry.client.v3.securitygroups.UnbindRunningSecurityGroupRequest;
 import org.cloudfoundry.client.v3.securitygroups.UnbindStagingSecurityGroupRequest;
 import org.cloudfoundry.client.v3.securitygroups.UpdateSecurityGroupRequest;
+import org.cloudfoundry.util.PaginationUtils;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.springframework.beans.factory.annotation.Autowired;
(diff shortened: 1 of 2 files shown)


Kehrlann (approved):

integration-test/src/test/java/org/cloudfoundry/client/v3/SecurityGroupsTest.java
```diff
@@ -196,6 +198,31 @@ public void listStaging() {
                 .verify(Duration.ofMinutes(5));
     }
 
+    @Test
+    public void listWithPagination() {
```
Kehrlann, line 202: I am unsure whether this test really showcases #1242 .

When I revert the change to have a v2 paginated request, and change the test to use `.resultsPerPage(1)`, it still passes on against CC API 3.191.0.

Does it fail on your machine?

SaifuddinMerchant: The failure I've seen is - results-per-page is a invalid parameter for V3 API.
This matches what I expect given the specs
https://v3-apidocs.cloudfoundry.org/version/3.191.0/index.html#list-security-groups

org.cloudfoundry.client.v3.ClientV3Exception: CF-UnprocessableEntity(10008): Unknown query parameter(s): 'results-per-page'. Valid parameters are: 'page', 'per_page', 'order_by', 'created_ats', 'updated_ats', 'guids', 'names', 'running_space_guids', 'staging_space_guids', 'globally_enabled_running', 'globally_enabled_staging'
	at org.cloudfoundry.reactor.util.ErrorPayloadMappers.lambda$null$2(ErrorPayloadMappers.java:62)
	
I haven't been able to run the integration test stand alone, but I do get the issue above using resultsPerPage(x) in my application

Kehrlann: Got it.

The test is incorrect, and is not testing what's intended, because when you do 

```java
this.securityGroup
        .map(...)
```

This returns a `Mono<Flux<SecurityGroupResource>>`. By subscribing to this, through `.verify(...)`, only the `this.securityGroup` request is ever called. The subsequent `Flux` is not called because nothing subscribes to it.

Instead, you should do:

```java
this.securityGroup
        .flux()
        .flatMap(/* ... */)
        .as(StepVerifier::create)
        // ...
```

This gives you a `Flux<SecurityGroupResource>`, and all calls are made. By doing that, I'm getting the error with the v2 pagination:

```
[ERROR] Tests run: 1, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 0.835 s <<< FAILURE! -- in org.cloudfoundry.client.v3.SecurityGroupsTest
[ERROR] org.cloudfoundry.client.v3.SecurityGroupsTest.listWithPagination -- Time elapsed: 0.830 s <<< FAILURE!
java.lang.AssertionError: expectation "expectNextCount(1)" failed (expected: count = 1; actual: counted = 0; signal: onError(org.cloudfoundry.client.v3.ClientV3Exception: CF-UnprocessableEntity(10008): Unknown query parameter(s): 'results-per-page'. Valid parameters are: 'page', 'per_page', 'order_by', 'created_ats', 'updated_ats', 'guids', 'names', 'running_space_guids', 'staging_space_guids', 'globally_enabled_running', 'globally_enabled_staging'))
```

SaifuddinMerchant: Greatly Appreciate the feedback and help, I will make the change

SaifuddinMerchant: I'm also going to try and see if I can run the actual integration test for this method on my cluster, so I might take some time to do that update

SaifuddinMerchant: I have fixed the test based on your feedback. No change from what you suggested. I can't run the integration test module but I was able to run just the specific test case with some modifications from my own module.

P.S I had based my test case on `public void list()` that I think suffers from a similar issue aka the listSecurityGroup is never called.

PR #1288, opened 2025-06-19
fix CVE‑2024‑58103

s-yonkov-yonkov:


Kehrlann:
Thanks for your contribution @s-yonkov-yonkov  !

Unfortunately, this is not a fix. We release a JDK-8 based jar only, with a Wire version of 3.x ; so this would only affect local builds. Happy to take this contribution in, but please reword the commit (something similar to the title of the issue as I updated it).

We will release a proper fix in the next 6.x release, see https://github.com/cloudfoundry/cf-java-client/pull/1262#issuecomment-2958382493

PR #1306, opened 2025-10-06
Move labels and annotations to metadata per CloudFoundry spec

xiaocongli:


Kehrlann:
Got it. I'm away this week. Will pick this up on Monday.

integration-test/src/test/java/org/cloudfoundry/operations/ApplicationsTest.java
```diff
@@ -789,6 +789,54 @@ public void pushManifestV3() throws IOException {
...
+                                                                                .build())
+                                                                .path(
+                                                                        new ClassPathResource(
+                                                                                        "test-application.zip")
+                                                                                .getFile()
+                                                                                .toPath())
+                                                                .build())
+                                                .build())
+                                .build())
+                .map(manifest -> manifest.getMetadata())
```
Lokowandtg, line 829: This line does not compile, as the "manifest" is of type "Void" and that type does not declare a "getMetadata" method.
If you are building with maven, please use this commandline:
./mvnw clean install -Dgpg.skip=true -Pintegration-test

cloudfoundry-operations/src/main/java/org/cloudfoundry/operations/applications/_ManifestV3Application.java
```diff
@@ -31,25 +30,17 @@
...
-
     /**
      * Generate a default route based on the application name
      */
     @Nullable
     abstract Boolean getDefaultRoute();
 
     /**
-     * The labels configured for this application
+     * The metadta for this application
```
Lokowandtg, line 40:
```suggestion
     * The metadata for this application
```

PR #1324, opened 2025-12-09
DefaultDomains: deprecate and point to client v3 api

Kehrlann:
This is part of moving `operations` to use the v3 API (#1325)

Since there is a single endpoint for domain management now, and no complex operations, I propose deprecating this API entirely, and remove it in 7.0


Kehrlann:
Superseded by #1327

PR #1330, opened 2025-12-17
Feature/apply organization and space quota

SaifuddinMerchant:


cloudfoundry-client/src/main/java/org/cloudfoundry/client/v3/quotas/spaces/_ApplySpaceQuotaRequest.java
```diff
@@ -0,0 +1,28 @@
...
+
+    /**
+     * The Space Quota id
+     */
+    @JsonIgnore
+    abstract String getSpaceQuotaId();
+
+    /**
+     * Relationships to the spaces where the quota is applied
+     * Use of JsonUnwrapped to inline the space relationships as per the API spec
```
Kehrlann, line 24: **[praise]** Thanks for the comment, very useful!

cloudfoundry-client/src/main/java/org/cloudfoundry/client/v3/quotas/spaces/_ApplySpaceQuotaResponse.java
```diff
@@ -0,0 +1,33 @@
...
+     */
+    @JsonUnwrapped
+    abstract ToManyRelationship spaceRelationships();
+
+    /**
+     * Links to related resources and actions for the resource
+     */
+    @AllowNulls
+    @JsonProperty("links")
+    public abstract Map<String, Link> getLinks();
```
Kehrlann, line 32: **[thought, non-blocking]** Perhaps this should got into a common base class. AFAICT, there are links on every single API response in v3.

SaifuddinMerchant: Yes, typically all the responses are able to extend from Resource that contains the getLinks and a few other properties. However Resource contains a few additional fields like createdAt, updatedAt that that ApplySpaceQuotaResponse does not have as it does not return a resource. It's kind of a outliner API :)

Issue #1325, opened 2025-12-10
Migrate Operations to use V3 API


Kehrlann:
To address CAPI v2 EOL (#1312), we need to migrate all Operations classes to use the v3 API. This work will induce breaking changes.

This issue tracks all the operations individually to follow what has been ported

## To port

- [ ] DefaultApplications
- [ ] DefaultServices
- [ ] DefaultOrganizationAdmin
- [ ] DefaultOrganizations.java
- [ ] DefaultServiceAdmin
- [ ] DefaultSpaceAdmin
- [ ] DefaultSpaces
- [ ] DefaultStacks
- [ ] DefaultUserAdmin
- [ ] DefaultNetworkPolicies
- [ ] DefaultRoutes

## Done

- [x] DefaultBuildPacks (#1323)
- [x] DefaultDomains (#1324 - deprecated in favor of `v3.Domains`)

## Delivered

None

Kehrlann:
cc @gberche-orange for visibility

PR #1331, opened 2025-12-17
DefaultApplications: use v3 API

Kehrlann:
This is a WiP

Part of #1325


Kehrlann:
We should use the v3 manifest endpoint to get the diskQuota and memoryLimit ; for now we don't because it doesn't exist. This means that if an app has 0 web processes, we'll get -1 for the values.

PR #1333, opened 2025-12-26
issue-1294 - implement user version 3 api

SaifuddinMerchant:


cloudfoundry-client/src/main/java/org/cloudfoundry/client/v3/users/_CreateUserRequest.java
```diff
@@ -0,0 +1,57 @@
...
+    /**
+     * Username of the user to be created. This can only be provided together with origin.
+     */
+    @JsonProperty("username")
+    @Nullable
+    abstract String getUsername();
+
+    /**
+     * Origin of the user to be created. This can only be provided together with username and cannot be uaa.
+     */
```
Kehrlann: What kind of error do we get from CAPI if username is provided with userId ; or without origin? What if we pass the `uaa` origin?

SaifuddinMerchant: java.lang.AssertionError: expectation "expectNextCount(1)" failed (expected: count = 1; actual: counted = 0; signal: onError(org.cloudfoundry.client.v3.ClientV3Exception: CF-UnprocessableEntity(10008): Origin cannot be provided with 'guid'))

Kehrlann: Awesome, thanks for trying this out 🙇

integration-test/src/test/java/org/cloudfoundry/client/v3/UsersTest.java
```diff
@@ -0,0 +1,155 @@
...
+
+    @Autowired private CloudFoundryClient cloudFoundryClient;
+
+    @Test
+    public void create() {
+        String userId = this.nameFactory.getUserId();
+
+        this.cloudFoundryClient
+                .usersV3()
+                .create(CreateUserRequest.builder().userId(userId).build())
```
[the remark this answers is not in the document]
SaifuddinMerchant: The test case did not work on my installation. I get

[ERROR] org.cloudfoundry.client.v3.UsersTest.createWithUsername -- Time elapsed: 0.240 s <<< FAILURE!
java.lang.AssertionError: expectation "expectNext(test-user-ov8s2)" failed (expected: onNext(test-user-ov8s2); actual: onError(org.cloudfoundry.client.v3.ClientV3Exception: **UnknownError(10001): An unknown error occurred.**))

  Suppressed: org.cloudfoundry.client.v3.ClientV3Exception: UnknownError(10001): An unknown error occurred.
                at org.cloudfoundry.reactor.util.ErrorPayloadMappers.lambda$null$2(ErrorPayloadMappers.java:62)
                Suppressed: The stacktrace has been enhanced by Reactor, refer to additional information below:
Assembly trace from producer [reactor.core.publisher.MonoFlatMap] :
        reactor.core.publisher.Mono.checkpoint(Mono.java:2212)
        org.cloudfoundry.reactor.client.v3.users.ReactorUsersV3.create(ReactorUsersV3.java:50)
Error has been observed at the following site(s):
        *__checkpoint() ? at org.cloudfoundry.reactor.client.v3.users.ReactorUsersV3.create(ReactorUsersV3.java:50)
Original Stack Trace:

I don't know why, I don't have access to the actual CF installation logs and the unknown error does not help :)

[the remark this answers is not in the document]
SaifuddinMerchant: > My understanding of the docs ([Create a user](https://v3-apidocs.cloudfoundry.org/version/3.208.0/index.html#create-a-user)):
> 
> * If CAPI property `cc.allow_user_creation_by_org_manager` is enabled, ...
>   
>   * Are you sure that's the case in your deployment?
> * ... a UAA user will be automatically created if it does not exist yet. ...
>   
>   * So this creates a user in the UAA DB ; it must not exist already.
> 
> This must be run as Admin or OrgManager, but I guess you were already doing that?

This only comes into picture when -
The UAA user will be only created when **username and origin** have been provided instead of a guid. Additionally origin must be different from **uaa**. Admins and OrgManagers can make use of the UAA user creation.

Yes I am running as admin, I don't know what the value of the property 'cc.allow_user_creation_by_org_manager` is set to in my installation

cloudfoundry-client/src/main/java/org/cloudfoundry/client/v3/users/_CreateUserRequest.java
SaifuddinMerchant: Is it ok to mark this conversation to resolved? We would still have the other one going for an integration test with username and origin ...

PR #1338, opened 2026-03-03
Add log cache log streaming support

ZPascal:
## Summary

Adds live log streaming via Log Cache — the Java equivalent of `cf tail --follow`.

Previously the only Log Cache read path was `logsRecent`, a single snapshot request. There was no way to continuously receive new log envelopes without polling manually. This change introduces a `logsTail` API across all three relevant modules.

---

## Motivation

Cloud Foundry dropped the legacy Loggregator Doppler streaming endpoint (`DopplerClient.stream()`) in Loggregator ≥ 107.0 (CFD ≥ 24.3 / TAS ≥ 4.0). The CF CLI replaced it with a polling loop over the Log Cache `/api/v1/read` endpoint — that is what `cf tail --follow` does today. This PR brings the same capability to Java consumers.

---

## Changes

<details>
<summary>Changes in detail</summary>

### `cloudfoundry-client` — new `TailLogsRequest` and `LogCacheClient.logsTail()`

New `_TailLogsRequest.java` (Immutables `@Value.Immutable`):

| Field | Type | Default | Description |
|---|---|---|---|
| `sourceId` | `String` | — | App / service GUID (**required**) |
| `startTime` | `Long` (nullable) | `now − 5 s` (ns) | Cursor start time in UNIX nanoseconds |
| `envelopeTypes` | `List<EnvelopeType>` (nullable) | all types | Envelope type filter |
| `nameFilter` | `String` (nullable) | none | Regex name filter (Log Cache ≥ 2.1.0) |
| `pollInterval` | `Duration` | `250 ms` | Back-off between polls when no new data |

New method on `LogCacheClient`:

```java
/**
 * Continuously polls Log Cache /api/v1/read and streams new Envelopes as they appear.
 * Equivalent to the Go logcache.Walk() API and `cf tail --follow`.
 * The Flux never completes on its own – cancel the subscription to stop streaming.
 */
Flux<Envelope> logsTail(TailLogsRequest request);
```

---

### `cloudfoundry-client-reactor` — non-blocking polling implementation

`ReactorLogCacheEndpoints.logsTail()` mirrors the Go [logcache.Walk()](https://github.com/cloudfoundry/go-log-cache/blob/main/walk.go) algorithm, fully non-blocking:

1. An `AtomicLong` cursor starts at `startTime` (or `now − 5 s`).
2. `Flux.defer` builds a fresh `ReadRequest` from the current cursor on every repetition and calls `GET /api/v1/read/{sourceId}?start_time=cursor`.
3. Envelopes are sorted ascending by timestamp; the cursor advances to `lastTimestamp + 1`; each envelope is emitted individually downstream.
4. When the batch is empty, `repeatWhen` inserts a `Mono.delay(pollInterval)` before the next poll. When envelopes arrive, the next poll starts immediately.
5. The `Flux` is infinite; the caller cancels the subscription to stop.

Transient network errors are logged and retried. HTTP errors (4xx/5xx) are propagated to the subscriber.

---

### `cloudfoundry-operations` — `Applications.logsTail()`

`DefaultApplications` delegates to `LogCacheClient`:

```java
@Override
public Flux<Envelope> logsTail(TailLogsRequest request) {
    return this.logCacheClient
            .flatMapMany(client -> client.logsTail(request))
            .transform(OperationsLogging.log("Tail Application Logs"))
            .checkpoint();
}
```

</details>

---

## Tests

| Test | What it verifies |
|---|---|
| `logsTailLogCache` | Happy path: a single `LOG/OUT` envelope is forwarded correctly |
| `logsTailLogCacheMultipleEnvelopes` | 3 envelopes with types `OUT → ERR → OUT` are emitted in order |
| `logsTailLogCacheError` | A `RuntimeException` from the client propagates unchanged to the subscriber |
| `logsTailLogCacheOutAndErrEnvelopes` | Both stdout and stderr envelopes are forwarded without filtering |

---

## Usage

```java
TailLogsRequest request = TailLogsRequest.builder()
        .sourceId(applicationGuid)
        .envelopeTypes(List.of(EnvelopeType.LOG))
        .pollInterval(Duration.ofMillis(250))
        .build();

logCacheClient.logsTail(request)
        .filter(e -> e.getLog() != null)
        .map(e -> e.getLog().getPayloadAsText())
        .subscribe(System.out::println);
```

Or via the Operations API:

```java
cloudFoundryOperations.applications()
        .logsTail(TailLogsRequest.builder().sourceId(appGuid).build())
        .filter(e -> e.getLog() != null)
        .map(e -> e.getLog().getPayloadAsText())
        .subscribe(System.out::println);
```

---

## Relation to existing API

| Method | Transport | Completes? | Use case |
|---|---|---|---|
| `DopplerClient.stream()` ⚠️ deprecated | WebSocket / Doppler | Yes (server closes) | Legacy streaming (Loggregator < 107.0) |
| `LogCacheClient.recentLogs()` | HTTP GET (single) | Yes | Fetch last N log lines |
| `LogCacheClient.logsTail()` ✅ new | HTTP GET (polling loop) | Never | Live streaming (Loggregator ≥ 107.0) |

---

## Checklist

- [x] `_TailLogsRequest` Immutables value object
- [x] `LogCacheClient.logsTail()` interface method
- [x] `ReactorLogCacheEndpoints.logsTail()` — non-blocking Reactor implementation
- [x] `_ReactorLogCacheClient.logsTail()` — delegate override
- [x] `Applications.logsTail()` — Operations API method
- [x] `DefaultApplications.logsTail()` — implementation
- [x] Unit tests in `DefaultApplicationsTest` (120 passing)
- [ ] Integration tests


ZPascal:
### Code review

Found 1 issue:

1. **`logs(ApplicationLogsRequest)` regression: `recent=true` no longer routes to Log Cache** — The PR replaces the pre-existing branch that routed `recent=true` to Log Cache (`getLogsLogCache`) with an unconditional delegation to the Doppler path (`logs(LogsRequest)`). On modern CF deployments (TAS 4.x / CFD 24.3+) where Doppler is unavailable, callers passing `recent=true` will silently fail. The existing `logsLogCache` unit test mocks `logCacheClient.read()` and verifies Log Cache routing for this case, but will no longer be exercised by this code path.

https://github.com/cloudfoundry/cf-java-client/blob/b0ce71fe6f9d3687723c114f9b289cebb8bc11c4/cloudfoundry-operations/src/main/java/org/cloudfoundry/operations/applications/DefaultApplications.java#L577-L595

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

cloudfoundry-client/src/main/java/org/cloudfoundry/logcache/v1/_TailLogsRequest.java
```diff
@@ -0,0 +1,65 @@
+/*
+ * Copyright 2013-2021 the original author or authors.
```
Lokowandtg, line 2:
```suggestion
 * Copyright 2026 the original author or authors.
```

cloudfoundry-client-reactor/src/main/java/org/cloudfoundry/reactor/logcache/v1/ReactorLogCacheEndpoints.java
```diff
@@ -48,4 +56,111 @@ Mono<MetaResponse> meta(MetaRequest request) {
...
+                                            .startTime(cursor.get());
+                            if (!envelopeTypes.isEmpty()) {
+                                builder.envelopeTypes(envelopeTypes);
+                            }
+                            if (nameFilter != null && !nameFilter.isEmpty()) {
+                                builder.nameFilter(nameFilter);
+                            }
+
+                            return read(builder.build())
+                                    .onErrorReturn(ReadResponse.builder().build())
```
Lokowandtg, line 119: This will drop any error without notice. For temporary hickups that is OK and wanted, but not for errors that persist (wrong credentials, wrong url,..., anything in the 4xx range).
At least an entry in the log should be written for non 4xx problems and if the problem will not vanish over time, the error should be returned.

integration-test/src/test/java/org/cloudfoundry/operations/ApplicationsTest.java
```diff
@@ -82,18 +88,18 @@
...
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Qualifier;
 import org.springframework.core.io.ClassPathResource;
 import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
+import reactor.core.publisher.SignalType;
 import reactor.test.StepVerifier;
 
 @CleanupCloudFoundryAfterClass
-@RequiresV2Api
```
Lokowandtg, line 96: This annotation was added lately. I do not see a reason to remove it. More comments below.

integration-test/src/test/java/org/cloudfoundry/operations/ApplicationsTest.java
```diff
@@ -356,7 +363,6 @@ public void getManifest() throws IOException {
     }
 
     @Test
-    @RequiresTcpRouting
```
Lokowandtg, line 359: see above

integration-test/src/test/java/org/cloudfoundry/operations/ApplicationsTest.java
```diff
@@ -443,7 +449,6 @@ public void getStopped() throws IOException {
...
```
Lokowandtg, line 446: see above

integration-test/src/test/java/org/cloudfoundry/operations/ApplicationsTest.java
```diff
@@ -1201,7 +1305,6 @@ public void pushRoutePath() throws IOException {
...
```
Lokowandtg, line 1204: see above

integration-test/src/test/java/org/cloudfoundry/operations/ApplicationsTest.java
```diff
@@ -1325,7 +1428,6 @@ public void pushUpdateRoute() throws IOException {
...
```
Lokowandtg, line 1328: see above

integration-test/src/test/java/org/cloudfoundry/operations/ApplicationsTest.java
```diff
@@ -753,59 +910,6 @@ public void pushManifestV3() throws IOException {
...
-                                        page ->
-                                                this.cloudFoundryClient
-                                                        .applicationsV3()
-                                                        .listFeatures(
-                                                                ListApplicationFeaturesRequest
-                                                                        .builder()
-                                                                        .applicationId(
-                                                                                applicationId)
-                                                                        .page(page)
-                                                                        .build())))
```
Lokowandtg, line 800: No need to delete a valid test.

Lokowandtg (changes requested):

cloudfoundry-client/src/main/java/org/cloudfoundry/logcache/v1/_TailLogsRequest.java
ZPascal: done

cloudfoundry-client-reactor/src/main/java/org/cloudfoundry/reactor/logcache/v1/ReactorLogCacheEndpoints.java
ZPascal: Done. I've also integrated the 5xx range.

integration-test/src/test/java/org/cloudfoundry/operations/ApplicationsTest.java
ZPascal: Done

integration-test/src/test/java/org/cloudfoundry/operations/ApplicationsTest.java
ZPascal: Done

integration-test/src/test/java/org/cloudfoundry/operations/ApplicationsTest.java
ZPascal: done

integration-test/src/test/java/org/cloudfoundry/operations/ApplicationsTest.java
ZPascal: done

integration-test/src/test/java/org/cloudfoundry/operations/ApplicationsTest.java
ZPascal: done

integration-test/src/test/java/org/cloudfoundry/operations/ApplicationsTest.java
ZPascal: done

PR #1339, opened 2026-03-12
Create/Update route options

theghost5800:
Refactor _RouteOptions class to handle added options in future

### DIFF
--- a/cloudfoundry-client/src/main/java/org/cloudfoundry/client/v3/routes/_RouteOptions.java
+++ b/cloudfoundry-client/src/main/java/org/cloudfoundry/client/v3/routes/_RouteOptions.java
@@ -1,8 +1,14 @@
 package org.cloudfoundry.client.v3.routes;
 
+import java.util.Map;
+import java.util.Optional;
+
+import com.fasterxml.jackson.annotation.JsonAnyGetter;
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
-import org.cloudfoundry.Nullable;
+import org.cloudfoundry.AllowNulls;
 import org.immutables.value.Value;
 
 /**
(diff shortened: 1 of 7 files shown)


theghost5800:
@Kehrlann Is it necessary something else to be added before code review to be done?

cloudfoundry-client/src/main/java/org/cloudfoundry/client/v3/routes/_RouteOptions.java
```diff
@@ -13,10 +19,40 @@
 abstract class _RouteOptions {
 
     /**
-     * The loadbalancing
+     * All route options, including unknown future keys.
```
Kehrlann, line 22: That's not a pattern we use anywhere, I don't think we should introduce it unless we have a very good reason to have it for this one specific sub-resource.

Otherwise, let's just add the hashes for now.

Kehrlann (changes requested):
Thanks for your contribution.
Could you please also add an integration test, even if you can't run it?

cloudfoundry-client/src/main/java/org/cloudfoundry/client/v3/routes/_RouteOptions.java
theghost5800: The reason why this pattern is used is because the client library doesn't keep up well with all new features introduced in Cloudfoundry API and it's annoying that consumers of cf-java-client library cannot use latest available features until it's contributed and released. So if there is an option to make some logic more generic, I think it's better for all users. What do you think?

Kehrlann: This is an interesting idea, but I don't want to add it on a class-by-class basis. Is there a way to add it to our Immutables and-or JSON config?

theghost5800: I have a look for some generic solution but it won't be easy to implement, so I go back to previous way of implementation by adding missing parameter options. I have added also integration test but I couldn't run it, so I hope that you can run and give back feedback is it okay.

theghost5800:
@Kehrlann Do you have any other comments? Is there any issues with new integration test run?

integration-test/src/test/java/org/cloudfoundry/CloudFoundryVersion.java
```diff
@@ -55,6 +55,8 @@ public enum CloudFoundryVersion {
 
     PCF_2_13(Version.forIntegers(2, 186, 0)),
 
+    PCF_2_14(Version.forIntegers(2, 248, 0)),
```
Lokowandtg, line 58: As we are dealing with an v3-API it would be more apropriate to use the version number 3.183.0, which is in the same capi-release.

Lokowandtg (approved):
The integration tests work and the code looks good.

integration-test/src/test/java/org/cloudfoundry/CloudFoundryVersion.java
Kehrlann: we don't have version checking for v3 so I'm fine with this.
With the environments we're running, I don't think the gate matters, they're all way ahead of 2.248 (CAPI 1.198)

Kehrlann:
@theghost5800 do you need this soon? Because if we target `main` it will only go in with version 6.0 and I have no timeline for that.

Kehrlann (approved):

theghost5800:
> @theghost5800 do you need this soon? Because if we target `main` it will only go in with version 6.0 and I have no timeline for that.

I am expecting this change for quite long time. So I am looking to use this feature with next release of the client library. If main branch won't be release anytime soon, please advice which branch must be targeted.

Lokowandtg:
Please target branch 5.x.x

theghost5800:
I have opened another PR to target branch 5.x.x - https://github.com/cloudfoundry/cf-java-client/pull/1365

Kehrlann:
Closed via acdbd2af13bd9d5168ddac9d78444ea1f444bcaa
This will be forward-ported to main

PR #1340, opened 2026-03-13
Support newer UAA response fields and grant types

jorbaum:
Different (likely newer?) versions of the open-source UAA (tested against v78.6.0) return fields and values that the client doesn't model, causing deserialization failures.

This PR adds support for:
  - `urn:ietf:params:oauth:grant-type:jwt-bearer` grant type (RFC 7523)
  - `defaultIdentityProvider` in identity zone configuration (added to [OSS UAA in 2018](https://github.com/cloudfoundry/uaa/commit/dac4722a51))
  - `defaultIdpName` in the /info endpoint response ([source](https://github.com/cloudfoundry/uaa/blob/284dd502db2316bfbdb05c47d9a6d23e3c82ba6a/server/src/main/java/org/cloudfoundry/identity/uaa/login/LoginInfoEndpoint.java#L277))
 
Integration test changes:

  - Relaxed assertion for UAA "access denied" error message wording
  - Check for login endpoint instead of password reset endpoint which might be disabled for some UAA


Kehrlann (dismissed):
LGTM, please target 5.x.x

Lokowandtg (dismissed):

jorbaum:
Rebased on 5.x.x and now targeting it.

PR #1342, opened 2026-03-18
make json parser more robust for unknown properties. #1341

### LINKED ISSUE #1341: Json Parser is too strict
The parser that handles the responses from the cloud foundry instance is very strict and aborts processing as soon as an unknown property is received or a type does not match. There is no easy way for applications using the cf-java-client to continue processing.
This becomes a problem if the versions of client and server do not match or if test coverage in the integration-tests does not point to the problem during development of a new version.

There should be some central switch to log and ignore problems during message parsing as far as possible. The switch would be helpfull during development, when client and server versions do not match.

There should be a configuration file to ignore only explicitly listed properties that are save to ignore, even in a productive environment.

Lokowandtg:
Fixes #1341 by improving the error handling when the JsonCodec detects a parsing error.
When there is a rule configured for the error at hand, the offending element is deleted and parsing of the remaining message is retried. If there is no rule or the retry fails, the problem is written to the log and the normal error handling stays unchanged.

An example configuration is in the integration tests. A configuration to ignore all errors is also provided.


Kehrlann:
If we ignore all the JSON parsing code, this specific fix is covered by #1340 I think, could you please confirm this?

Lokowandtg:
https://github.com/cloudfoundry/cf-java-client/pull/1340 fixes the missing features we currently face because cf-java-client is behind.

This change is improving error handling, making it configurable. This allows future incompatibilities to be ignored without touching the code.

So both changes should be merged.

Kehrlann:
@Lokowandtg have you seen my comments in #1341 ? ([direct link](https://github.com/cloudfoundry/cf-java-client/issues/1341#issuecomment-4125091864))

Lokowandtg:
Hi Daniel,
yes I have seen your comment and was a bit frustrated that I wasted time to implement the ~400 lines. 😊

The topic is not so pressing since I have a working test environment where I can control the version to use and when to update.

I will come up with a simpler switch / env variable for the integration tests when I find the time. By default, the integration tests should fail when there is a change. When the variable is set, they should continue working on best effort.

Thanks, Georg

PR #1343, opened 2026-03-23
Add non v2 log cache read only test

jorbaum:
Add a new test to verify log cache read works in an env without V2.

Only the latests commit is relevant.


jorbaum:
Sorry.. did not want to open a PR for this one..

Issue #1344, opened 2026-03-23
Integration tests fail: test-service-broker.jar incompatible with java-buildpack v5 (JDK 17 default)


jorbaum:
On March 20-21, 2026, the java-buildpack [`main` branch was updated to v5.0.1](https://github.com/cloudfoundry/java-buildpack/pull/1217), which [defaults to JDK 17](https://github.com/cloudfoundry/java-buildpack/blob/main/manifest.yml) (up from JDK 8 in v4.x).

The integration tests hardcode `buildpack("https://github.com/cloudfoundry/java-buildpack.git")` in [`ApplicationUtils.java:230`](https://github.com/cloudfoundry/cf-java-client/blob/main/integration-test/src/test/java/org/cloudfoundry/ApplicationUtils.java#L230), so they now get JDK 17. The bundled [`test-service-broker.jar`](https://github.com/cloudfoundry/cf-java-client/blob/main/integration-test/src/test/resources/test-service-broker.jar) is a Spring Boot 1.5.16 app (built ~2018) that relies on CGLIB reflective access [incompatible with JDK 17's module system](https://github.com/cglib/cglib/issues/191). The app crashes on startup with `InaccessibleObjectException`, causing `DelayTimeoutException` and failing all tests that depend on `serviceBrokerId`.

### Workaround

Add `--add-opens` flags to the service broker environment in [`ServiceBrokerUtils.pushServiceBrokerApplication()`](https://github.com/cloudfoundry/cf-java-client/blob/main/integration-test/src/test/java/org/cloudfoundry/ServiceBrokerUtils.java#L145):

```java
env.put("JAVA_OPTS",
    "--add-opens java.base/java.lang=ALL-UNNAMED"
        + " --add-opens java.base/java.io=ALL-UNNAMED");
```

Potential longer-term fixes

- Rebuild test-service-broker.jar with a modern Spring Boot version
- Pin the buildpack to a version tag or use the system buildpack instead of tracking main

Kehrlann:
We should also rebuild `test-log-cache.jar` as part of this effort too.

PR #1345, opened 2026-03-23
Fix test-service-broker crash on JDK 17 by adding --add-opens flags

jorbaum:
Workaround for #1344 .

The test-service-broker.jar (Spring Boot 1.5.16 / Spring Framework 4.3.19) uses CGLIB proxying which requires reflective access to java.lang.ClassLoader.defineClass(). JDK 17's module system blocks this, causing the app to crash on startup with
InaccessibleObjectException and all ApplicationsTest methods to fail with DelayTimeoutException during serviceBrokerId bean creation.

Creating this as a draft as this is a workaround only to some parts of the integration tests work again.

### DIFF
--- a/integration-test/src/test/java/org/cloudfoundry/ServiceBrokerUtils.java
+++ b/integration-test/src/test/java/org/cloudfoundry/ServiceBrokerUtils.java
@@ -145,6 +145,10 @@ public final class ServiceBrokerUtils {
         Map<String, Object> env = new HashMap<>();
         env.put("SERVICE_NAME", serviceName);
         env.put("PLAN_NAME", planName);
+        env.put(
+                "JAVA_OPTS",
+                "--add-opens java.base/java.lang=ALL-UNNAMED"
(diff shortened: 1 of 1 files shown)


Kehrlann (dismissed):
Thanks for the hotfix. Please target the 5.x.x branch. We'll forward port this.

jorbaum:
I pushed the version that was rebased on 5.x.x .

However, there is also [test-log-cache.jar](https://github.com/cloudfoundry/cf-java-client/blob/631746446f5266284130fc3a7b081466eb4b96e8/integration-test/src/test/java/org/cloudfoundry/IntegrationTestConfiguration.java#L609) which might have similar issues. Even though I did not run into issues with it for the integration tests that I ran.

Do you think the changes in this PR are fine without looking into / adding an additional workaround for test-log-cache.jar ?

Kehrlann:
The logcache client app seems to be a more recent build (Boot 2.x). I think it's less likely to be affected.
Let's leave it as-is, and if it fails, we'll apply the same fix. Thanks!

PR #1346, opened 2026-03-23
Support skipping integration tests for unavailable CF features via environment variables

jorbaum:
On my CF test environment quite a few integration tests needed features that my environment did not support.

I added a bunch of environment variables so those tests can be disabled at a global level.

I am not sure about this approach yet. Especially regarding disabling V2 CAPI as it touches quite a lot of places. Any feedback welcome!


integration-test/src/test/java/org/cloudfoundry/client/v3/ServiceBindingsTest.java
```diff
@@ -54,11 +55,14 @@
...
 @IfCloudFoundryVersion(greaterThanOrEqualTo = CloudFoundryVersion.PCF_2_11)
+@RequiresV2Api
 public class ServiceBindingsTest extends AbstractIntegrationTest {
 
     @Autowired private CloudFoundryClient cloudFoundryClient;
 
-    @Autowired private Mono<String> serviceBrokerId;
+    // Optional: bean requires V2 API; class is guarded by @RequiresV2Api
+    @Autowired(required = false)
+    private Mono<String> serviceBrokerId;
```
Kehrlann, line 65: Isn't it surprising that the class pulls in the `serviceBrokerId` bean when it's full disabled? I would have expected that this bean is not loaded if the class doesn't execute.

integration-test/src/test/java/org/cloudfoundry/logcache/v1/LogCacheTest.java
```diff
@@ -68,6 +83,7 @@ public void info() {
     }
 
     @Test
+    @RequiresV2Api
```
Kehrlann, line 86: Only the `info` test can execute without the V2 API, and it doesn't bring much value. Let's annotate the whole class .

integration-test/src/test/java/org/cloudfoundry/CloudFoundryCleaner.java
```diff
@@ -129,6 +130,12 @@ final class CloudFoundryCleaner implements InitializingBean, DisposableBean {
 
     private static final Logger LOGGER = LoggerFactory.getLogger("cloudfoundry-client.test");
 
+    private static final boolean RUN_V2_CLEANUP = isRunV2Tests();
+
+    private static boolean isRunV2Tests() {
+        return !"true".equalsIgnoreCase(System.getenv("SKIP_V2_TESTS"));
+    }
```
Kehrlann, line 137: Consider putting this in `RequiresV2Api` so the condition is shared.

integration-test/src/test/java/org/cloudfoundry/CloudFoundryCleaner.java
```diff
@@ -241,32 +275,43 @@ void clean() {
...
+                                application ->
+                                        nameFactory.isApplicationName(application.getName()));
+
+        if (runV2Calls) {
+            apps =
+                    apps.delayUntil(
+                            application ->
+                                    removeApplicationServiceBindings(
+                                            cloudFoundryClient, application));
+        }
```
Kehrlann, line 298: That's the only v2 calls, which lists then deletes service bindings.
Can we port it to v3 instead of gating it with a boolean?

integration-test/src/test/java/org/cloudfoundry/IntegrationTestConfiguration.java
```diff
@@ -350,6 +351,7 @@ RandomNameFactory nameFactory() {
 
     @Bean(initMethod = "block")
     @DependsOn("cloudFoundryCleaner")
+    @ConditionalOnProperty(name = "SKIP_V2_TESTS", havingValue = "false", matchIfMissing = true)
```
Kehrlann, line 354: Let's centralize this property name, e.g. in `RequiresV2Api`.

README.md
```diff
@@ -298,6 +298,10 @@ Name | Description
 `TEST_PROXY_USERNAME` | _(Optional)_ The username for a proxy to route all requests through
 `TEST_SKIPSSLVALIDATION` | _(Optional)_ Whether to skip SSL validation when connecting to the Cloud Foundry instance.  Defaults to `false`.
 `UAA_API_REQUEST_LIMIT` | _(Optional)_ If your UAA server does rate limiting and returns 429 errors, set this variable to the smallest limit configured there. Whether your server limits UAA calls is shown in the log, together with the location of the configuration file on the server. Defaults to `0` (no limit).
+`SKIP_BROWSER_AUTH_TESTS` | _(Optional)_ Set to `true` to skip integration tests that require browser-based authentication (e.g., SSO tests). Useful when the Cloud Foundry instance does not support browser-based authentication or when running tests in a non-interactive environment. Defaults to `false`.
```
Kehrlann, line 301: Is this really about "non-interactive" environments? Do we support interactive environments in our tests? Or is it about some UAA configuration that has browser-based auth?

Kehrlann (commented):
- Really clear annotations for execution conditions, I like the approach.
- A few questions and remarks throughout the PR.
- Please target `5.x.x`

My biggest issue is that half of the classes in the v3 package are annotated with `@RequiresV2Api`. It seems v2 is required because we push applications (service broker, most notably) in the config with the v2 API, and those tests needs application pushed.
Could you please consider updating the core integration test config to use v3 apis for pushing apps instead? 
To have examples of this, look at the `ApplicationsTest` in the v3 package.

jorbaum:
One big reason for why I wanted to create a draf PR initially (which, apparently I did not :P ) is that I only tested the new annotations in the following combination:

```bash
SKIP_V2_TESTS=false \
SKIP_TCP_ROUTING_TESTS=true \
SKIP_BROWSER_AUTH_TESTS=true \
SKIP_METRIC_REGISTRAR_TESTS=true
```

and 

```bash
SKIP_V2_TESTS=true \
SKIP_TCP_ROUTING_TESTS=true \
SKIP_BROWSER_AUTH_TESTS=true \
SKIP_METRIC_REGISTRAR_TESTS=true
```

This is because my test environment does not support tcp routing, browser-based auth or cf marketplace. Do you think this is fine?

Kehrlann:
I think it's fine. I'll test locally before merging, with and without the flags.

please also target 5.x.x, we'll merge this back into main later.

integration-test/src/test/java/org/cloudfoundry/client/v3/ServiceBindingsTest.java
jorbaum: Good question. AFAIU this is because `Mono.block()` is called in [IntegrationTestConfiguration.java](https://github.com/cloudfoundry/cf-java-client/blob/a8f7c0925307942884851ee17477db6249b28b1a/integration-test/src/test/java/org/cloudfoundry/IntegrationTestConfiguration.java#L475). The JUnit annotation just disables the class at test execution time.

Instead of doing `@Autowired(required = false)` I used `@ConditionalOnProperty` in IntegrationTestConfiguration now.

README.md
jorbaum: Sorry. You are right. The env vars purpose is to skip tests that require specific UAA configuration for browser-based auth. However, completing them programmatically or not should not matter here.

jorbaum:
> Could you please consider updating the core integration test config to use v3 apis for pushing apps instead?

I have now addressed all of the comments except for this one. Moving the test config to use v3 seems like a bigger endeavor. I did not try to tackle it so far.

Also I did not manage to rerun the integration tests again. Feel free to take a look at my changes already though.

jorbaum:
I now ran this against https://github.com/Lokowandtg/kind-deployment/tree/cf-java-client.integration-test and found a bunch of issues that I fixed with the latest commits.

In those I migrated organizationI, spaceId, stackId and stackName from V2 to V3. But refrained from migrating app pushing so far as its a larger endeavor. Also note that the default quota is now used for the organization.

I also found a Json deserialization issue in Stack.java that I fixed.

I ran the following integration tests:

* org.cloudfoundry.networking.v1.*Test
* org.cloudfoundry.client.v3.*Test
* org.cloudfoundry.logcache.v1.*Test

Kehrlann:
Currently running whole integration test suite.

Issue #1263, opened 2025-04-03
Add support for V3 API app-features


beyhan:
The CF API V3 defines [app-features](https://v3-apidocs.cloudfoundry.org/version/3.190.0/index.html#app-features) resource, which gets more and more adoption. E.g. the [RFC-0030](https://github.com/cloudfoundry/community/blob/main/toc/rfc/rfc-0030-add-support-for-file-based-service-binding.md) is making use of this but there are also other use cases. That is why, it will be great to support this API in cf-java-client.

PR #1351, opened 2026-04-13
5.x.x app features

aegliv:
Implements https://github.com/cloudfoundry/cf-java-client/issues/1263

### DIFF
--- a/cloudfoundry-operations/src/main/java/org/cloudfoundry/operations/applications/_ApplicationManifestCommon.java
+++ b/cloudfoundry-operations/src/main/java/org/cloudfoundry/operations/applications/_ApplicationManifestCommon.java
@@ -46,10 +46,6 @@ abstract class _ApplicationManifestCommon {
         Builder environmentVariable(Map.Entry<String, ? extends Object> entry);
         Builder environmentVariables(@Nullable Map<String, ? extends Object> entries);
         Builder putAllEnvironmentVariables(Map<String, ? extends Object> entries);
-        Builder feature(String key, Object value);
-        Builder feature(Map.Entry<String, ? extends Object> entry);
-        Builder features(@Nullable Map<String, ? extends Object> entries);
-        Builder putAllFeatures(Map<String, ? extends Object> entries);
         Builder healthCheckHttpEndpoint(@Nullable String healthCheckHttpEndpoint);
         Builder healthCheckType(@Nullable ApplicationHealthCheck healthCheckType);
         Builder host(String element);
@@ -145,13 +141,6 @@ abstract class _ApplicationManifestCommon {
     @AllowNulls
     @Nullable
     abstract Map<String, Object> getEnvironmentVariables();
-
-    /**
-     * Manage whether optional capabilities are enabled
-     */
-    @AllowNulls
-    @Nullable
-    abstract Map<String, Object> getFeatures();
 
     /**
      * The HTTP health check endpoint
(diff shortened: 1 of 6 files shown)


aegliv:
@Kehrlann ping

Kehrlann:
(I'm currently off sick, will run the integration test next week.)

integration-test/src/test/java/org/cloudfoundry/operations/ApplicationsTest.java
```diff
@@ -16,68 +16,15 @@
...
-import org.cloudfoundry.operations.applications.ScaleApplicationRequest;
-import org.cloudfoundry.operations.applications.SetApplicationHealthCheckRequest;
-import org.cloudfoundry.operations.applications.SetEnvironmentVariableApplicationRequest;
-import org.cloudfoundry.operations.applications.StartApplicationRequest;
-import org.cloudfoundry.operations.applications.StopApplicationRequest;
-import org.cloudfoundry.operations.applications.Task;
-import org.cloudfoundry.operations.applications.TaskState;
-import org.cloudfoundry.operations.applications.TerminateApplicationTaskRequest;
-import org.cloudfoundry.operations.applications.UnsetEnvironmentVariableApplicationRequest;
+import org.cloudfoundry.operations.applications.*;
```
Kehrlann: The imports are broken, please revert to explicit imports and just add what you need.

integration-test/src/test/java/org/cloudfoundry/operations/ApplicationsTest.java
```diff
@@ -573,6 +530,54 @@ public void pushBindServices() throws IOException {
...
+                                                                .build())
+                                                .noStart(false)
+                                                .build()))
+                .then(
+                        this.cloudFoundryOperations
+                                .applications()
+                                .getApplicationManifest(
+                                        GetApplicationManifestRequest.builder()
+                                                .name(applicationName)
+                                                .build()))
```
Kehrlann: This does not return ApplicationFeatures (it's null), because this `getApplicationManifest` is a synthetic manifest build by CFJC itself ([source](https://github.com/cloudfoundry/cf-java-client/blob/main/cloudfoundry-operations/src/main/java/org/cloudfoundry/operations/applications/DefaultApplications.java#L407-L440))

You want to either:
1. Update operations to get the app features and put them in the manifest
2. Update the test to use the v3 features API to list features and check the result.

Kehrlann (changes requested):

integration-test/src/test/java/org/cloudfoundry/operations/ApplicationsTest.java
aegliv: Plz try again

cloudfoundry-operations/src/main/java/org/cloudfoundry/operations/applications/_ApplicationManifestCommon.java
```diff
@@ -142,6 +146,13 @@ void check() {
     @Nullable
     abstract Map<String, Object> getEnvironmentVariables();
 
+    /**
+     * Manage whether optional capabilities are enabled
+     */
+    @AllowNulls
+    @Nullable
+    abstract Map<String, Object> getFeatures();
```
Kehrlann: features don't seem to work with manifest v2 at all.

Only add it to manifest v3

integration-test/src/test/java/org/cloudfoundry/operations/ApplicationsTest.java
```diff
@@ -573,6 +576,65 @@ public void pushBindServices() throws IOException {
                 .verify(Duration.ofMinutes(5));
     }
 
+    @Test
+    public void pushAppFeatures() throws IOException {
```
Kehrlann: The tests do not break anymore.

However, the features are not pushed, because .pushApplicationManifest v2 does not actually push a manifest, it does a few manual push app operations. You need to use manifest v3 only.

Kehrlann (changes requested):

integration-test/src/test/java/org/cloudfoundry/operations/ApplicationsTest.java
aegliv: done

aegliv:
@Kehrlann Did you have time to look at it again?

Kehrlann:
Thanks for your contribution, this works.

The integration test doesn't run if you use a V3 version number, though. Will enable it in a follow-up commit: 9426f1ed2a2c6d3770d5942b41dc0545345c86b5

PR #1354, opened 2026-04-17
Cherry-pick changes from 5.x.x into main

jorbaum:
I cherry-picked all the commits after https://github.com/cloudfoundry/cf-java-client/commit/6c81106cb75faf2bd30eebd5232fbc9469eff6ab and excluded the SNAPSHOT commit.

PR #1355, opened 2026-04-23
Forward merge 5.x.x into main

jorbaum:
Brings all changes from 5.x.x into main. Reverts version bump commit.

Compared to https://github.com/cloudfoundry/cf-java-client/pull/1354 the "Fix client credentials grant token invalidation" commit will appear twice in the history. Also, the diff on github shows changes coming from this commit, but git should be able to sort this out.


Kehrlann (approved):
This is perfect, thanks!

Indeed, since I cherry-picked the `Fix client_credentials` on 5.x.x from main, the commit appears in both branches of the graph, but git can do the merge without making any change. I tried a local merge and it went fine.

I also did further forward-ports of 5.x.x into main and it goes smoothly.

Issue #1229, opened 2024-04-15
v2/info call getting triggered internally while creating CloudFoundryClient and consuming v3 api's


bhavinshah-mendix:
Hello,

Could someone assist me in grasping why the invocation of v3 API functions such as organizationsV3(), spacesV3(), or routesV3() etc., triggers the automatic invocation of v2/info when I create a CloudFoundryOperation client using CloudFoundryClient?

**CloudFoundryClient Creation code:**
```
cloudFoundryClient = ReactorCloudFoundryClient.builder()
        		.connectionContext(connectionContext)
                .tokenProvider(tokenProvider)
                .build();
```
**OperationClient Creation Code:**
```
cloudFoundryOperations = DefaultCloudFoundryOperations.builder()
                .cloudFoundryClient(cloudFoundryClient)
                .organization(StringUtils.EMPTY)
                .space(StringUtils.EMPTY)
                .build();

```
**Fetching Organization:**
```
ListOrganizationsRequest orgListRequest = ListOrganizationsRequest.builder().build();
ListOrganizationsResponse listOrganizationsResponse =  cloudFoundryClient.organizationsV3().list(orgListRequest).timeout(CF_REQUEST_TIMEOUT).block();
```

After investigation, I discovered that within _ReactorCloudFoundryClient.java, there's a checkCompatibility() method annotated with @PostConstruct, triggering the info() function. However, it remains uncertain whether this is the sole factor contributing to the observed behaviour.

looping @anthonydahanne , as I noticed your impressive contribution to the repository! Your dedication and expertise shine through in your work, so could help me :)

Lokowandtg:
I will try to replace the internal usages of /v2/info with corresponding calls in /3/info and in "/". Here is the mapping for calls I have found:
```
/v2/info.name						/v3/info.name
/v2/info.build						/v3/info.build
/v2/info.support					/v3/info.links.support.href
/v2/info.version					/v3/info.version
/v2/info.description				  	/v3/info.description
/v2/info.authorization_endpoint				/.links.login.href
/v2/info.token_endpoint					/.links.uaa.href
/v2/info.min_cli_version				/v3/info.cli_version.minimum
/v2/info.min_recommended_cli_version			/v3/info.cli_version.recommended 
/v2/info.app_ssh_endpoint				/.links.app_ssh.href
/v2/info.app_ssh_host_key_fingerprint			/.links.app_ssh.meta.host_key-fingerprint
[...]

PR #1356, opened 2026-04-28
make compatibilityChecker robust if v2 api is disabled. dirty fix for #1229

Lokowandtg:
With this change, the compatibility checker does not fail if v2 API is not available.
In that case, no check is executed, which is not correct.
I will work on adding v3/info endpoint and checking that version next.


Kehrlann:
Hey @Lokowandtg 

Questions:
1. What happens today we have a foundation with v2 disabled? What's the error?
2. You plan to add v3 compatibility, but API versions are different between v2 and v3. What's your plan?

Lokowandtg:
Without this pull request, the CloudFoundryClientCompatibilityChecker runs into a NullPointerException when v2 API is not available.
The next plan is to check all "v3" integration-tests where some "v2" endpoint is called and see if there is a replacement that uses only v3. So v2-integration-tests should stay as is, v3 should only use v3 and the rest I will check later.

Kehrlann:
@Lokowandtg this looks good.

Could you please:
- target 5.x.x
- add an integration test for the new Root ?

Kehrlann:
Merged through #1359

PR #1364, opened 2026-06-10
Minor fixes to some integration tests. fixes #1363

### LINKED ISSUE #1363: Minor bugs in integration-tests
The "InfoTest" fails if running against a CF where v2 API is disabled. 
.v3.OrganizationsTest.getDefaultDomain fails when running in kind-deployment landscape.

The package "serviceInstances" (with capital "I") does not match the package declaration in the contained class. This creates problems in Eclipse on Windows.

Pull request will follow.
https://github.com/cloudfoundry/cf-java-client/pull/1364

Lokowandtg:
InfoTest fails if running against a CF where v2 API is disabled. 
OrganizationsTest.getDefaultDomain fails when running in kind-deployment landscape.

The package "serviceInstances" (with capital "I") does not match the package declaration in the contained class. This creates problems in Eclipse on Windows.


integration-test/src/test/java/org/cloudfoundry/client/v2/SpacesTest.java
```diff
@@ -1885,13 +1885,9 @@ public void listServicesFilterByLabel() {
                 .verify(Duration.ofMinutes(5));
     }
 
-    // TODO: Await https://github.com/cloudfoundry/cloud_controller_ng/issues/856 for this test to
-    // work
-    @Disabled(
-            "Await https://github.com/cloudfoundry/cloud_controller_ng/issues/856 for this test to"
-                    + " work")
```
Kehrlann, line 1892: Awesome!

integration-test/src/test/java/org/cloudfoundry/client/v3/OrganizationsTest.java
```diff
@@ -183,7 +183,21 @@ public void getDefaultDomain() {
...
+                                                    assertThat(nameParam)
+                                                            .contains(
+                                                                    "apps.",
+                                                                    ".shepherd.tanzu.broadcom.net"),
+                                            nameParam ->
+                                                    assertThat(nameParam)
+                                                            .contains(
+                                                                    "apps.",
+                                                                    ".127-0-0-1.nip.io")); // when
+                            // testing with kind-deploy.
```
Kehrlann: Suggestion: Can we infer the domain from the `TEST_APIHOST` env var?

(Not a hard requirement but it'd be nicer than hardcoding deployment options)

integration-test/src/test/java/org/cloudfoundry/client/v3/PackagesTest.java
```diff
@@ -30,17 +30,16 @@
...
 import org.cloudfoundry.client.v3.packages.UploadPackageRequest;
-import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Test;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.core.io.ClassPathResource;
 import reactor.core.Exceptions;
 import reactor.core.publisher.Mono;
 import reactor.test.StepVerifier;
 
-@Disabled("Until Packages are no longer experimental")
```
Kehrlann, line 43: Very nice!

integration-test/src/test/java/org/cloudfoundry/client/v2/InfoTest.java
```diff
@@ -43,8 +43,15 @@ public void info() {
...
+                            Version actual;
+                            String version = response.getApiVersion();
+                            if (version == null || version.isEmpty()) {
+                                assertThat("CF API v2 is disabled")
+                                        .isEqualTo(response.getSupport());
+                                actual = Version.of(0, 0, 0);
+                            } else {
+                                actual = Version.valueOf(version);
+                            }
                             assertThat(actual).isLessThanOrEqualTo(expected);
```
Kehrlann, line 55: This file is annotated with `@RequiresV2Api`. Have you tried the `SKIP_V2_TESTS` [env var](https://github.com/cloudfoundry/cf-java-client/tree/5.x.x#development)? Does it not work as intended?

I'm surprised other test annotated with `@RequiresV2Api` passed at all.

Kehrlann (commented):
Hey @Lokowandtg 

Thanks for your contribution. Surprised that some tests (all the `.client.v2` and `.operations` tests) pass without V2, are you using `SKIP_V2_TESTS` ? And if so, shouldn't `InfoTest` be skipped?

integration-test/src/test/java/org/cloudfoundry/client/v3/OrganizationsTest.java
Lokowandtg: That's a good point. Will do.

integration-test/src/test/java/org/cloudfoundry/client/v2/InfoTest.java
Lokowandtg: None of the tests pass when v2 is disabled completely. So far v2 is fully functional, but the urls are checked and v2 is refused. This refusal code is disabled locally :-)
Current plan is to remove v2 starting 2027, by then we must be done with the migration.

jorbaum: AFAIU the whole `org.cloudfoundry.client.v2` package requires CAPI v2 API.

Yes, it should be skipped when setting `SKIP_V2_TESTS` to true.

Lokowandtg: SKIP_V2_TESTS works as intended, but I ususally set it to "false" and run all tests.

integration-test/src/test/java/org/cloudfoundry/client/v3/OrganizationsTest.java
Lokowandtg: Done.

Kehrlann: Doesn't work on my setup:


```
[ERROR]   Run 1: OrganizationsTest.lambda$getDefaultDomain$7:189
Expecting actual:
  "apps.tas.b225eb9d.shepherd.tanzu.broadcom.net"
to contain:
  ["apps.", ".sys.tas.b225eb9d.shepherd.tanzu.broadcom.net"]
but could not find:
  [".sys.tas.b225eb9d.shepherd.tanzu.broadcom.net"]
```

Lokowandtg: Fixed. As there is no way to derive the default domain from the ApiHost, I just check that the endpoint returns something.

PR #1366, opened 2026-07-17
issue-1357: Add attribute allowpublic to uaa Client resource

damzog:


Lokowandtg:
As far as I can see, the code looks good and the existing tests run.
But there are no integration-tests. Can you add some or extend an existing one so it passes/validates the "allowpublic" value? If you need a local test environment, you can use [kind-deployment](https://github.com/Lokowandtg/kind-deployment/tree/cf-java-client.integration-test). If that is a problem, please tell us where (in which existing test) the new flag will be passed to CF landscape and can be checked.
Thanks, Georg

damzog:
@Lokowandtg I have added explicit setting and validating the flag in an existing integration test.

I actually tried to run the integration test against a kind-deployment (setting it up actually caused me some trouble) but I could not figure out the required values for TEST_ADMIN_CLIENTID , TEST_ADMIN_CLIENTSECRET ( I guess a client needed to be created first which would require to know uaa admin user name ) maybe if you can share some light into this?

Anyway the code is so simple it might even run without having tested it ...

Lokowandtg:
Running the integration test in my landscape fails, as there are two clients created, only the second one having the "allowpublic" set, but the first client is used for validation and results in a null value.
As I have no way to push to your fork, here is a fixed version of the test:
`    @Test
    public void batchCreate() {
        String clientId1 = this.nameFactory.getClientId();
        String clientId2 = this.nameFactory.getClientId();
        String clientSecret = this.nameFactory.getClientSecret();

        this.uaaClient
                .clients()
                .batchCreate(
                        BatchCreateClientsRequest.builder()
                                .client(
                                        CreateClient.builder()
                                                .approvalsDeleted(true)
                                                .authorizedGrantType(PASSWORD)
                                                .clientId(clientId1)
                                                .clientSecret(clientSecret)
                                                .allowPublic(false)
                                                .scopes("client.read", "client.write")
                                                .tokenSalt("test-token-salt")
                                                .build())
                                .client(
                                        CreateClient.builder()
                                                .approvalsDeleted(true)
                                                .authorizedGrantTypes(PASSWORD, REFRESH_TOKEN)
                                                .clientId(clientId2)
                                                .clientSecret(clientSecret)
                                                .allowPublic(true)
                                                .scope("client.write")
                                                .tokenSalt("filtered-test-token-salt")
                                                .build())
                                .build())
                .flatMapIterable(BatchCreateClientsResponse::getClients)
                .collectMap(Client::getClientId)
                .as(StepVerifier::create)
                .consumeNextWith(
                        clients -> {
                            Client client1 = clients.get(clientId1);
                            assertThat(client1).isNotNull();
                            assertThat(client1.getAuthorizedGrantTypes())
                                    .containsExactly(PASSWORD, REFRESH_TOKEN);
                            assertThat(client1.getScopes())
                                    .containsExactly("client.read", "client.write");
                            assertThat(client1.getTokenSalt()).isEqualTo("test-token-salt");
                            assertThat(client1.getAllowPublic()).isFalse();

                            Client client2 = clients.get(clientId2);
                            assertThat(client2).isNotNull();
                            assertThat(client2.getAuthorizedGrantTypes())
                                    .containsExactlyInAnyOrder(PASSWORD, REFRESH_TOKEN);
                            assertThat(client2.getScopes()).containsExactly("client.write");
                            assertThat(client2.getTokenSalt())
                                    .isEqualTo("filtered-test-token-salt");
                            assertThat(client2.getAllowPublic()).isTrue();
                        })
                .expectComplete()
    }
`

Lokowandtg:
Do you know when (in which version of CAPI) "allowpublic" was added? A guard-annotation might be needed.

damzog:
@Lokowandtg  that is actually embarrassing. I will change the test case as you propose.

The allowPublic attribute (allowing the client_secret to be omitted when public clients use PKCE with code_challenge_method=S256) was introduced in UAA version 75.21.0., see [release notes](https://github.com/cloudfoundry/uaa/releases?page=10#release-v75.21.0l)

Lokowandtg (approved):

Kehrlann (approved):

Kehrlann:
Thanks for your contribution!

PR #1369, opened 2026-07-27
Add env-var to ignore unknown properties. Fix #1341

### LINKED ISSUE #1341: Json Parser is too strict

Lokowandtg:
Setting "TEST_FAIL_ON_UNKNOWN_PROPERTIES" to "false" will ignore unknown properties returned from the CF instance because of version mismatch. Default is "true", so we notice incompatibilities early. Developers can switch to "false" to continue working.


Kehrlann (approved):

Kehrlann:
Very useful, thanks!

